PR #30, opened 2012-09-10
Adding formatted output for -Q as per issue #19.

techhat:
As it says.


s0undt3ch:
Hmm, salt-cloud requires salt to as a dependency?
If it does, it could make use of salt's parser.

@thatch45 want me to have a look at it?

thatch45:
Yes, salt-cloud requires salt, and yes, we could add it to the parser :)

PR #32, opened 2012-09-11
write /etc/salt before installing salt-minion on Ubuntu deploy.

lexual:
This fix for #25 means we don't need to restart the minion at all.

### DIFF
--- a/saltcloud/deploy/Ubuntu.sh
+++ b/saltcloud/deploy/Ubuntu.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
+mkdir -p /etc/salt/pki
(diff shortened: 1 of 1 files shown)


thatch45:
Just to double check, apt does not throw a file conflict error here? I may just be getting the little points of package managers mixed up, but I would think that apt would get upset because it is trying to replace an existing file, /etc/salt/minion

lexual:
"When a package is installed for the first time dpkg will install the file that comes with it, unless that would mean overwriting a file already on the file system."

from http://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html

thatch45:
Thanks for the conformation!

PR #44, opened 2012-09-17
Filter out stale (terminated) AWS instances.

dangarthwaite:


thatch45:
Awesome, thanks @dangarthwaite

PR #45, opened 2012-09-18
EC2 profile and security group updates

mafrosis:
Two fixes:
- allow profiles to override core keyname & securitygroup props
- allow lists of security groups on EC2 instance


thatch45:
Very nice! Thanks Matt!

PR #47, opened 2012-09-19
Do not always expect master_finger or grains

techhat:
A previous commit had a side effect of always requiring these values to be set. This commit makes them optional.


thatch45:
remind me in the morning to show you an easier way for these

PR #48, opened 2012-09-19
Show helpful error when a bad aws location is used

techhat:
Seriously, I wasted too much time because this was missing.


akoumjian:
I know it's a tiny change, but don't we normally want someone other then the author to merge the pull request?

thatch45:
It is ok if the author merges small changes.

akoumjian:
That's fair.

techhat:
Normally, I would agree... but as Tom said, it was a tiny change.

thatch45:
You are right though, and for the vast majority of them they should be reviewed

PR #49, opened 2012-09-19
Switch linode from using libcloud deploy to use salt-cloud wait_for_ssh

techhat:
libcloud's deploy isn't working so great, so we're switching to their create + our wait_for_ssh


thatch45:
Looks great! Thanks!

PR #50, opened 2012-09-20
Switch rackspace from libcloud deploy to salt-cloud deploy

techhat:
This series of commits successfully switches rackspace from a libcloud deploy, to using our own in-house methods. This is a big step towards implementing our own standard deploy method.


thatch45:
w00t!

PR #82, opened 2012-10-09
Have setup.py properly pull in requirements

techhat:
Right now this is mostly important for Paramiko, but will probably be important for other deps in the future.


techhat:
Addresses issue #71.

PR #86, opened 2012-10-11
Query (-Q or -F) based on provided map (-m)

techhat:
Addresses issue #64.


techhat:
Oops, accidentally added the fix for issue #63 in here as well. But since it uses some of the code from #64, it might as well be in here anyway.

PR #97, opened 2012-10-20
Added barebones deploy script for SmartOS

techhat:
This probably needs to be fleshed out better, but I believe the steps themselves to work.


thatch45:
You are the man!

PR #107, opened 2012-10-25
Only define class Cloud in cloud.py

ConsoleCatzirl:
There is no reason to have to separate definitions of class Cloud


thatch45:
Heh, yes, seems I never got this cleaned up :)

PR #110, opened 2012-10-29
Merge hpcs provider into openstack provider

ConsoleCatzirl:
The openstack and hpcs providers where developed in parallel.  This pulls effort from hpcs into the more general openstack provider.


saltcloud/libcloudfuncs.py
```diff
@@ -103,6 +104,7 @@ def get_size(conn, vm_):
             return size
         if str(size.name) == str(vm_['size']):
             return size
+    raise ValueError
```
whiteinge: These `raise` calls should have a helpful error message, such as `raise ValueError("The foobar is missing baz")`.

techhat:
This is good stuff, thanks man!

PR #111, opened 2012-10-29
Changes to the Ubuntu deploy scripts

ConsoleCatzirl:
This changes the shell for the Ubuntu deploy scripts from `bash` to `sudo bash`, and also starts the salt-minion service once it is installed.


techhat:
Looks good, thanks duder!

PR #122, opened 2012-11-09
Fix deleting nodes from a map file

ConsoleCatzirl:
This fixes deleting nodes from a map file when not all nodes have grains specified.

Previously, if a node in the list was simply a string, an uncaught exception was thrown.
Now, a list of lists is created from the mix of strings and dict keys, then flattened into a simple list.


thatch45:
Very nice! Thanks!

PR #132, opened 2012-11-15
Pyflakes

mafrosis:
I found a handful of bugs in utils/**init**.py when trying to debug and fix another bug to do with SSH usernames (another P/R coming ;))

All these were obvious when running pyflakes in Vim - would it be possible to start doing so before committing? I wasted a lot of time being confused due to 0b0b338 because the exception there meant that the time.sleep() call was happening in the outer Exception handler...

Cheers

### DIFF
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -305,6 +305,7 @@ def root_cmd(command, tty, sudo, key_filename, username, host):
                 )
         subprocess.call(cmd, shell=True)
     else:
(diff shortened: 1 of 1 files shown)


thatch45:
Yes, we have a jenkins server that checks these in Salt, I want to add one for the other projects

techhat:
Ick, I'm sorry you had to deal with this, @mafrosis. It looks like you beat me to cleaning up a few things that I had found (and caused), and then found some that I didn't know about yet. I'll have to do some serious regression tests before the next release.

PR #133, opened 2012-11-15
wait_for_passwd logic fix

mafrosis:
The paramiko.SSHClient connect() method doesn't actually return a boolean, causing my deploy to always fail:

http://www.lag.net/paramiko/docs/paramiko.client-pysrc.html

I'm not entirely sure how anything could deploy correctly with the code as it was - and as I recall most testing doesn't happen against EC2.

Could this fix be tested to ensure it doesn't break for other providers?

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -213,12 +213,10 @@ def create(vm_):
     else:
         ip_address = data.public_ips[0]
     if saltcloud.utils.wait_for_ssh(ip_address):
-        username = 'ec2-user'
         for user in usernames:
             if saltcloud.utils.wait_for_passwd(host=ip_address, username=user, timeout=60, key_filename=__opts__['AWS.private_key']):
                 username = user
                 break
-        kwargs['ssh_username'] = username
     deploy_command='bash /tmp/deploy.sh'
     if username == 'root':
         deploy_command='/tmp/deploy.sh'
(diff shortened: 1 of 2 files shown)


techhat:
Actually, most of my testing is with EC2, because we're using a trial account and it's currently the cheapest (^_^).

The two lines in aws.py are there for a reason. Amazon Linux defaults to ec2-user, and those two lines are set up to provide that user as a default. At the time I believed it to be correct, but I could be wrong. Unless you have a specific reason for removing those two lines, I'd like to keep them in there.

I appreciate your logic fix in wait_for_passwd. It drove me crazy all day yesterday, and I only ran a few tests before calling it good.

mafrosis:
No worries :)

The default of "ec2-user" is already setup elsewhere, is it not? I was checking out the iteration over usernames, and in my case it looked like:
['ubuntu', 'ec2-user', 'root']

So I presumed that was some superfluous debug or something. Either way, the username variable and the kwargs variable aren't used after that!

I think the wait_for_passwd() fix is good though.

techhat:
I'm only too happy to merge the wait_for_passwd() fix.

Looking closely, username is used after that, when calling deploy_script, but kwargs doesn't seem to be. However, you're right, ec2-user should already be in that list. I will go ahead and merge your change.

mafrosis:
Oh sorry, yeah I remember seeing username is passed to deploy_script(). And in that case, it wouldn't work for my setup where the login user is "ubuntu".

The current code should work though, where username will be whatever was successful in wait_for_passwd().

techhat:
That's right, you're using the ubuntu images. I'm tempted to add ubuntu to the list of default usernames, but having two defaults is already kind of a long list... and really, most users on AWS should be specifying the correct name anyway. I kind of hope nobody ever really has to get as far as trying defaults.

PR #134, opened 2012-11-15
Option to use private ip for deploying with openstack

balboah:
Some environments will only set a private ip.
Use same option name as AWS


techhat:
This is good stuff, thanks man!

PR #135, opened 2012-11-15
Don't force a name for openstack compute endpoint

balboah:
This is optional and there should be a configuration option for it if we're about to force it.


techhat:
I like this, this is better than what I was about to commit.

ConsoleCatzirl:
Sorry to be a stickler, but can we update docstring when we update options?  Now that the service_name isn't forced to a hardcoded value, we should add an example setting to the yaml block in the docstring.

balboah:
I usually complain on documentation but didn't update it myself so yeah I will add that ;)

PR #143, opened 2012-11-16
Fix Rackspace stacktrace error

techhat:
Rackspace seems to have updated their API recently. Copying some code from the OpenStack module is a temporary fix, until we can get apikey support going in OpenStack.

While troubleshooting this issue, I went ahead and removed some unnecessary prints, and added/updated logging statements where appropriate.

### DIFF
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -20,7 +20,6 @@ log = logging.getLogger(__name__)
 try:
     import paramiko
 except:
-    print('Cannot import paramiko. Please make sure it is correctly installed.')
     log.error('Cannot import paramiko. Please make sure it is correctly installed.')
     sys.exit(1)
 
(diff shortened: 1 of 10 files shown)


saltcloud/utils/__init__.py
```diff
@@ -268,43 +270,53 @@ def deploy_script(host, port=22, timeout=900, username='root',
...
         # Tried this with paramiko's invoke_shell(), and got tired of
         # fighting with it
+        log.debug('Using ssh command to simulate tty session, to execute command')
         cmd = ('ssh -oStrictHostKeyChecking=no -t -i {0} {1}@{2} "{3}"').format(
-                key_filename,
-                username,
-                host,
+                kwargs['key_filename'],
+                kwargs['username'],
+                kwargs['hostname'],
```
ConsoleCatzirl, line 311: By changing these to kwargs, they are no longer required, so we should probably check that they are set before using them.

thatch45:
Yes, @jesusaurus is right, these should be verified

PR #159, opened 2012-11-21
Better checking for minion config

techhat:
A master should always be defined, for any minion being deployed.


techhat:
This is related to issue #116.

PR #165, opened 2012-11-26
More explicit import error message. Minor PEP-8 fixes.

s0undt3ch:


techhat:
@s0undt3ch thanks for prettying up my ugly version check.

PR #172, opened 2012-11-28
AWS Sydney region support. Requires patch for https://issues.apache.org/...

lexual:
Here's a patch that enables support for the new AWS region in Sydney (ap-southeast-2).

http://aws.typepad.com/aws/2012/11/asia-pacific-sydney-region-open.html

It requires the patch for this libcloud bug I filed, requesting AWS Sydney support. I believe this is currently only in trunk, not in a release.

https://issues.apache.org/jira/browse/LIBCLOUD-264

https://github.com/lexual/salt-cloud/commit/c1c5189b7ec420d62764fed80bf054ece87a4fa4


techhat:
This is such a simple change, that I think I'll accept it, even though it's not in an official upstream release yet. I know that @mafrosis might be interested in this as well, and possibly willing to update to trunk to get it.

PR #173, opened 2012-11-28
Break out sftp into its own function

techhat:
This function requires an existing transport, so that it doesn't have the overhead of having to connect and disconnect each time.

I have also added the ability to pass in minion.pem, minion.pub and minion (conf), which will be uploaded to /tmp/ (and chmodded in the case of minion.pem), so that they are available for deploy.sh. After deploy.sh finishes, they will also be automatically cleaned up.

This paves the way for bootstrap-salt-minion.sh to replace the existing deploy scripts in their entirety. This will also simplify user-end configuration.


techhat:
https://github.com/saltstack/salty-vagrant/blob/develop/scripts/bootstrap-salt-minion.sh is the new deploy script in question.

PR #174, opened 2012-11-29
Test for EC2_AP_SOUTHEAST2 attribute before adding it to the locations dictionary.

s0undt3ch:
Test for EC2_AP_SOUTHEAST2 attribute before adding it to the locations dictionary.


techhat:
I like it.

PR #175, opened 2012-11-29
check if vm in running state

runyaga:
without these changes i could not reliably provision vm's on rackspace.


techhat:
I'm not seeing any breakage on my side, so I'll go ahead and merge.

PR #177, opened 2012-11-29
In -Q (but not -S or -F) show actual state, not just the state code

techhat:
Because I'm sick of just seeing a number.


thatch45:
Are those values always consistent with libcloud?

techhat:
Yes, they can be found in the NodeState class of:

https://github.com/apache/libcloud/blob/trunk/libcloud/compute/types.py

PR #179, opened 2012-11-30
Reuse established connections

ConsoleCatzirl:
Allow for established connections to be passed in to various util functions to avoid needless re-authentication.


techhat:
This is definitely a step in the right direction. Having to connect and reconnect all the time is messy.

PR #180, opened 2012-11-30
We are using human-readable node states

ConsoleCatzirl:
Correctly check for a RUNNING state in the Openstack provider.


techhat:
Sorry, I didn't realize I caused this bug. Thanks for the fix.

PR #182, opened 2012-12-03
Passing "None" argument to quiet down weird error

techhat:
This started happening because of pull req #179:

  File "/usr/bin/salt-cloud", line 14, in <module>
    main()
  File "/usr/bin/salt-cloud", line 10, in main
    cloud.run()
  File "/usr/lib/python2.7/site-packages/saltcloud/cli.py", line 72, in run
    mapper.size_list(self.options.list_sizes)
  File "/usr/lib/python2.7/site-packages/saltcloud/cloud.py", line 132, in size_list
    sizes[prov] = self.clouds[fun]()
TypeError: avail_sizes() takes exactly 1 argument (0 given)

The weird thing is, avail_sizes() calls for conn=None, which shouldn't make it required. But passing in None seems to fix it, at least for the moment.

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -110,7 +110,7 @@ class Cloud(object):
                 # The capability to gather images is not supported by this
                 # cloud module
                 continue
-            images[prov] = self.clouds[fun]()
+            images[prov] = self.clouds[fun](None)
         return images
 
     def size_list(self, lookup='all'):
@@ -129,7 +129,7 @@ class Cloud(object):
                 # The capability to gather sizes is not supported by this
                 # cloud module
                 continue
-            sizes[prov] = self.clouds[fun]()
+            sizes[prov] = self.clouds[fun](None)
         return sizes
 
     def create_all(self):


adevore:
I'm a friend of jesusaurus. I looked into this by request.

avail_sizes interacts with a call to types.FunctionType. FunctionType acts on the __code__ attribute of avail_sizes. The values for the default argument for functions are attached to the function objects, not the __code__ objects. Because __code__ doesn't have any way to know what the default arguments are, it blows up.

For a demonstration, run this gist: https://gist.github.com/5f94c786b918a38afc63 . Note the contents of __defaults__ on the function object.

To avoid the problem, FunctionType needs an extra argument. The function signature for FunctionType is:

```
function(code, globals[, name[, argdefs[, closure]]])
```

I haven't tested this, but I think the solution is to pass the value of avail_sizes.__defaults__ in for argdefs.

__defaults__ is a Python 3 renaming of the earlier func_defaults. Python versions >= 2.6 have __defaults__ for forward compatibility, so you must use func_defaults if you want to support < 2.6.

s0undt3ch:
@adevore Thanks, I found this out as @techhat told me you posted this here.

techhat:
I like @s0undt3ch 's fix in #183 better.

whiteinge:
@adevore very helpful explanation! thank you

PR #185, opened 2012-12-03
Change vm to VM where appropriate

techhat:
Seriously, it's been driving my crazy, and my OCD won't let it continue. This aggression will not stand, man.


thatch45:
bwahaha!
Thanks @techhat !

PR #190, opened 2012-12-06
Only wait for minion to check if a start_action is specified

techhat:
I don't know how many people (if any) are even using start_action yet, but at least this will narrow the problem to just them... and a temporary fix for them would be to use startup states:

http://docs.saltstack.org/en/latest/ref/states/startup.html


techhat:
This should resolve issue #187.

PR #203, opened 2012-12-24
add repos directly, start/stop

ghost:
The script "add-apt-repository" is a convenience for listing a PPA's URL and fetching it's key, which can be done manually for about the same expence without the extra dep.. Also, add-apt-repository comes from different packages in different versions of ubuntu.

objections welcome! :)


techhat:
Oh man. We totally needed Debian.

PR #209, opened 2013-01-09
update default cloud conf file with aws ssh_interface example

inthecloud247:


techhat:
I like it. Thanks, @ydavid365!

PR #224, opened 2013-01-17
added mako support to salt-cloud map files.

inthecloud247:
if mako is installed, salt-cloud will take advantage of it by default and try to render map files as mako first and then output as yaml.

should probably extend to work with a config file option renderer='mako'. wanted to submit this before heading off for the day...

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -17,6 +17,8 @@ import salt.utils
 
 # Import third party libs
 import yaml
+try:
+    from mako.template import Template
 
 # Get logging started
 log = logging.getLogger(__name__)
@@ -302,7 +304,13 @@ class Map(Cloud):
             raise ValueError('The specified map file does not exist: {0}\n'.format(self.opts['map']))
         try:
             with open(self.opts['map'], 'rb') as fp_:
-                map_ = yaml.load(fp_.read())
+                try:
+                    #open mako file
+                    temp_ = Template(open(fp_, 'r').read()) 
+                    #render as yaml
+                    map_ = temp_.render() 
+                except:
+                    map_ = yaml.load(fp_.read())
         except Exception:
             return {}
         if 'include' in map_:


techhat:
Looks pretty straightforward. Thanks @ydavid365!

thatch45:
Woah, we should totally bootstrap the render system for this, not piece meal in random template languages

inthecloud247:
agreed... :-)

Would be a good idea to get the regular salt render system working here... had to just get mako working for now.

Also want to get basic pillar functionality in salt-cloud. but at least with mako now installed I can hack pillar support in by loading the pillar files using pyyaml... :-)

thatch45:
What is the application you see for pillar here?

inthecloud247:
I want to have pillars as the central 'source of truth' for any global data for my salt/salt-cloud cluster. So my nodes should be defined in pillars and then loaded into salt-cloud. I was able to hack this to work using mako code in my salt-cloud .map file. But it's not super robust, and ideally it should use the same shared salt template processing code. 

Here's my use case... I want to have a staging environment that I can easily bring up and down at will to test/validate our code and salt configuration.

By adding mako template code to our .map file, and by hacking in the ability to load pillar data into salt-cloud, I was able to easily build a staging environment in salt-cloud with the following machines:

#api machines
apple-staging.mydomain.com
banana-staging.mydomain.com
cherry-staging.mydomain.com
date-staging.mydomain.com

#load balancer machines
agnon-staging.mydomain.com
beckett-staging.mydomain.com

## In a servers.sls pillar file in pillar/staging/servers.sls

domain: staging.mydomain.com
environment: staging

api:
  role: 'api'
  size: 'small'
  tenant: 'single'
  dns: 'dynamic'
  instances:
    - {number: 2, provider: 'aws', region: 'west-2b'}
    - {number: 2, provider: 'aws', region: 'west-2c'}

lb:
  role: 'lb'
  size: 'small'
  tenant: 'single'
  dns: 'static'
  instances: 
    - {number: 1, provider: 'aws', region: 'west-2b'}
    - {number: 1, provider: 'aws', region: 'west-2c'}

lb_static_ips:
- 192.168.1.12
- 192.168.1.13

#and in a shared pillar: /pillar/servernamelists.sls

api_names:
- apple
- banana
- cherry
- date
- etrog

lb_names:
- agnon
- beckett
- camus
- deledda
- elytis

techhat:
@ydavid365, I'm not sure if I've pointed this to you yet or not, but you'll want to take a look at #220, which I hope to get done next week, after the 0.8.4 release.

inthecloud247:
@techhat that's really cool. Didn't know about that. seems like it would cover my use case really well... :-)

PR #225, opened 2013-01-17
Report syntax errors in map file. Exit when nothing to do for a map.

mafrosis:
As described, just a couple of small fixes relating to map files.

You may well want to change the "Nothing to do" message!

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -122,6 +122,9 @@ class SaltCloud(parsers.SaltCloudParser):
             self.exit(0)
 
         if self.options.map and self.selected_query_option is None:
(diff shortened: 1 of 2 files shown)


techhat:
@mafrosis, you'll want to do a git fetch and redo this commit.

mafrosis:
@techhat snap.. someone merged whilst I was committing!

I've fixed in my dev version - but there's an error in there. The new mako import has a broken try except on line 20. D'you want to fix this before I raise another P/R?

mafrosis:
Also this is another case for pylint!

techhat:
Fixed. I guess I got lazy.

PR #227, opened 2013-01-17
Report syntax errors in map file. Exit when nothing to do for a map.

mafrosis:
As before..


techhat:
Ah, much better. Thanks, duder!

PR #232, opened 2013-01-17
noinput on ubuntu install, keep user's config

akoumjian:
UNTESTED Quick fix for saltstack/salt-cloud#230

Should work until bootstrap script gets integrated in 8.4


techhat:
I'm going to merge this, since we will not be removing these scripts for several releases... but salt-bootstrap will be the default in the next release.

ghost:
thank you much @akoumjian ! good catch there!

PR #236, opened 2013-01-17
I heart shebangs

ghost:
per issue #20

### DIFF
--- a/saltcloud/deploy/Arch-git.sh
+++ b/saltcloud/deploy/Arch-git.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 echo '[salt]
 Server = http://red45.org/archlinux
--- a/saltcloud/deploy/Arch.sh
+++ b/saltcloud/deploy/Arch.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 echo '[salt]
 Server = http://red45.org/archlinux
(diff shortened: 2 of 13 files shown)


techhat:
Nope, sorry, can't merge this one. These are all distros that ship with Bash, and should be using it.

ghost:
I don't mean to take a hard line, but let me share what I know FWIW..

The LSB says "shell scripts must avoid using bash extensions if they want to run on any LSB system." [Linux Standards Base]

All sane shell scripts use only bourne syntax and POSIX 1003.2 --that's the API for shell scripts. Bash syntax is not part of that standard (though /bin/sh is actually provided by the bash package on most systems).  

None of the deploy scripts actually use any bash syntax and they all end in .sh (not ".bash").  Most importantly though, AFAIK these scripts are not directly executed but are 'sourced' for their functionality, so there is no control at all over which interpreter runs them from what is inside the script, and in that case they don't even need a shebang [they are all installed without execute permission]

Thanks for listening. :)

techhat:
They are executed directly, not sourced.

Normally, this would be a moot point anyway, since we are phasing out these scripts. However, they are being kept around for academic purposes (and because our users would balk if their scripts suddenly disappeared).

That said, I would prefer to keep bash at the top of the script. It has far more functionality than supposedly-sane scripts, and I don't want users seeing sh and assuming that there's some special reason why they can't use bash if they wanted to. I believe that the .sh extension is valid, because that's what users expect to see on both sh and bash scripts (but not necessarily ksh or csh).

The FreeBSD scripts don't use bash, because it can't be expected to be there on a FreeBSD system, which would qualify as a special reason why a user assume that sh is what they're stuck with. The other distros listed can currently be expected in have bash available.

s0undt3ch:
Ppl, sorry for the open/close noise. Once more.

PR #243, opened 2013-01-18
Add generic wrappers for salt-bootstrap

techhat:
This gives another method for the user to add their own commands before or after the deployment, without using startup states.

### DIFF
--- a/saltcloud/deploy/curl-bootstrap-git.sh
+++ b/saltcloud/deploy/curl-bootstrap-git.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# This is a generic wrapper for the salt-bootstrap script at:
+#
+# https://github.com/saltstack/salt-bootstrap
+# 
+# It has been designed as an example, to be customized for your own needs.
+
+curl -L http://bootstrap.saltstack.org | sudo sh -s git develop
--- a/saltcloud/deploy/curl-bootstrap.sh
+++ b/saltcloud/deploy/curl-bootstrap.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# This is a generic wrapper for the salt-bootstrap script at:
+#
+# https://github.com/saltstack/salt-bootstrap
+# 
+# It has been designed as an example, to be customized for your own needs.
+
+curl -L http://bootstrap.saltstack.org | sudo sh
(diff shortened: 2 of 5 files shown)


techhat:
I suspect @seanchannel will appreciate the shabangs here.

ghost:
would there be any reason against having the existing deploy scripts use the salt-bootstrap 'one-liner' as the default?

techhat:
Yes, which is unfortunate, because in some ways I would rather do that than ship a static copy. Some issues with using a one-liner are:
- Which one to use? The examples I provided (lifted straight from the salt-bootstrap README) use curl, wget and Python.
- curl and wget are out, because neither of them are part of the base install for most distros, and can't reasonably be expected to be installed on a fresh image.
- Python is also out, because while it is core to almost every Linux distro out there, it can't be expected to be installed on any of the Unices.

Then there's problems with not shipping a static salt-bootstrap. I'm going to be ripping the download back out of setup.py, because having it downloaded fresh every single time runs the risk of having a different version installed for every package. Worse, the differing versions could contain fundamental differences that cause problems in troubleshooting RedHat vs Debian based packages (for instance). Shipping one, known version may introduce upstream bugs from salt-bootstrap that are a pain to deal with, but at least they'll be consistent bugs.

inthecloud247:
well one solution would be to allow us to pin a salt-bootstrap version...

why not build in the ability to specify in our salt-cloud config the url for the salt-bootstrap script, the command to download the script (curl, wget, etc.) and the command to execute the script (python, sh, etc.) ? 
then we can have the defaults cover the most common cases (curl + python), and for more advanced users or users installing salt-bootstrap on machines lacking common utilities/programs they will have the flexibility to just handle it in whatever way they choose. For example they could modify salt-bootstrap to first install curl and python...

inthecloud247:
I have an annoying situation where I need to now have mako installed with salt, and it would be great if I could add 'yum install python-mako' to a `pre-bootstrap.sh` script so I don't need to do my current semi-hacky workaround using salt startup_states.

techhat:
I added these scripts to do exactly what you're talking about. Add "yum install python-mako" before the line that calls salt-bootstrap, and you're set.

inthecloud247:
Awesome! I realized that was a possibility last night when I started digging more into the documentation for salt bootstrap. But wasn't 100% sure... Thanks!

PR #244, opened 2013-01-18
Have setup.py download the latest salt-bootstrap automagically

techhat:
Addresses #149.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,17 @@ The setup script for salt
 '''
 
 import os
+import urllib
 from distutils.core import setup
+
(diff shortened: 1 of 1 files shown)


akoumjian:
Why are we doing this instead of using git submodules? This seems like a really unreliable way to deploy the bootstrap script. Would you want some control over what versions of bootstrap were attached to which versions of salt-cloud?

That way you also avoid all the silliness with curl vs. wget, since the script is already there. Lastly, why is there a separate script just for installing from git? Shouldn't the install options be part of the profile or cloud config and then appended to whatever executes it on the minion?

PR #249, opened 2013-01-19
fixed a few typos. documenting the new-style bootstrap os config option

inthecloud247:


techhat:
The os config option was changed in 0.8.2:

http://salt-cloud.readthedocs.org/en/latest/topics/releases/0.8.2.html

All of the documentation should have subsequently been updated to no longer reference the os option. If you could back out the non-typo changes, I do have docs already set up to explain the rest. I'll push them out on Monday.

inthecloud247:
ok cool! :-)

PR #252, opened 2013-01-21
Support for multiple regions in AWS

mafrosis:
Changes:
- added cli location parameter (-L or --location)
- initialize libcloud functions with a region-specific connection object
- modified namespaced_function to set default parameter when wrapping methods from libcloudfuncs

This patch should allow the region to be specified in the following order (each overriding the former):
1. Global salt-cloud config
2. Cloud profile setting
3. CLI parameter

Only AWS is supported in this patch.

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -48,15 +48,6 @@ from salt.exceptions import SaltException
 # Get logging started
 log = logging.getLogger(__name__)
 
-# Init the libcloud functions
-avail_images = namespaced_function(avail_images, globals())
-avail_sizes = namespaced_function(avail_sizes, globals())
-script = namespaced_function(script, globals())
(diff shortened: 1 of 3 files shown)


techhat:
You are my hero, @mafrosis!

inthecloud247:
@mafrosis The patch works great. One thing that still needs fixing is to have the duplicate name check only check if there is an instance running with the same name that isn't in 'state:terminated' state. Right now, if running `salt-cloud -Q` shows 'terminated' instances, you will not be able to create any new instances with those names until they disappear from the amazon api and disappear from the `salt-cloud -Q` results. Terminated instances should be not considered to exist for purposed of preventing duplicate machines.

``` yaml
    date-test.domain.net:
        id:
            i-dc4346ee
        image:
            None
        private_ips:
        public_ips:
        size:
            None
        state:
            TERMINATED
```

PR #256, opened 2013-01-21
ignore the auto-downloaded bootstrap file.

inthecloud247:


techhat:
Salt Cloud doesn't auto-download the bootstrap file anymore.

inthecloud247:
k cool. :-)

so we should just delete the bootstrap file then?

techhat:
Yup. The release tarball will have it in the correct place.

PR #258, opened 2013-01-21
confirmation for destroying nodes fixes #219

mafrosis:


techhat:
I'm not sure if this is in the right place... it doesn't look like it will affect -a destroy. And I'm 99% certain we had confirmation somewhere before for deleting VMs (especiall for maps). I am not sure, however, that we had confirmation for -a destroy.

mafrosis:
Basically this patch moves all confirmation into the cli module, and expands to add confirmation to both destroy and action modes.

FYI the message removed in 18bd9e8 is covered by the confirmation print in cli.py. Further to this, the delete_map method should probably be renamed.. I just didn't know what to call it ;)

techhat:
This is good stuff. Thanks, @mafrosis!

PR #261, opened 2013-01-22
fixing ubuntu.sh deploys...

inthecloud247:
basically the same as the ubuntu-git.sh deploy fix


techhat:
Let me run some tests on this guy.

inthecloud247:
updating these 2 into one pull request. made another fix...

PR #263, opened 2013-01-22
updated ubuntu and ubuntu-git deploy scripts. hands-free on 12.04-1

inthecloud247:
ok git snafus hopefully fixed. this should now work... :-)

### DIFF
--- a/saltcloud/deploy/Ubuntu.sh
+++ b/saltcloud/deploy/Ubuntu.sh
@@ -5,8 +5,9 @@ echo '{{ vm['priv_key'] }}' > /etc/salt/pki/minion.pem
 echo '{{ vm['pub_key'] }}' > /etc/salt/pki/minion.pub
 echo "{{ minion }}" > /etc/salt/minion
 
-echo deb http://ppa.launchpad.net/saltstack/salt/ubuntu `lsb_release -sc` main | tee /etc/apt/sources.list.d/saltstack.list
-wget -q -O- "http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0x4759FA960E27C0A6" | apt-key add -
+apt-get install -y python-software-properties
+#Add simulated [enter] key to add-apt-repository
+echo | add-apt-repository -y ppa:saltstack/salt
 apt-get update
 
 apt-get install -y -o DPkg::Options::=--force-confold salt-minion
--- a/saltcloud/deploy/Ubuntu-git.sh
+++ b/saltcloud/deploy/Ubuntu-git.sh
@@ -5,8 +5,9 @@ echo '{{ vm['priv_key'] }}' > /etc/salt/pki/minion.pem
 echo '{{ vm['pub_key'] }}' > /etc/salt/pki/minion.pub
 echo "{{ minion }}" > /etc/salt/minion
 
-echo deb http://ppa.launchpad.net/saltstack/salt/ubuntu `lsb_release -sc` main | tee /etc/apt/sources.list.d/saltstack.list
-wget -q -O- "http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0x4759FA960E27C0A6" | apt-key add -
+apt-get install -y python-software-properties
+#Add simulated [enter] key to add-apt-repository
+echo | add-apt-repository -y ppa:saltstack/salt
 apt-get update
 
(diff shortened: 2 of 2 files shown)


techhat:
Much better. Let me spin up a VM and check it out.

saltcloud/deploy/Ubuntu.sh
```diff
@@ -5,8 +5,9 @@ echo '{{ vm['priv_key'] }}' > /etc/salt/pki/minion.pem
 echo '{{ vm['pub_key'] }}' > /etc/salt/pki/minion.pub
 echo "{{ minion }}" > /etc/salt/minion
 
-echo deb http://ppa.launchpad.net/saltstack/salt/ubuntu `lsb_release -sc` main | tee /etc/apt/sources.list.d/saltstack.list
-wget -q -O- "http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0x4759FA960E27C0A6" | apt-key add -
+apt-get install -y python-software-properties
+#Add simulated [enter] key to add-apt-repository
+echo | add-apt-repository -y ppa:saltstack/salt
```
s0undt3ch, line 10: If you pass `-y` there's not need for the simulated enter.

s0undt3ch, line 10: Do note that `-y` is only supported for Ubuntu >= 10.04

inthecloud247, line 10: @s0undt3ch  even with -y, I still needed to simulate enter on ubuntu 12.04-1 . It's a known issue that sometimes it requires an enter. Pretty weird.

Well luckily salt-bootstrap also only supports Ubuntu >= 10.04, so we should be ok there I hope.

s0undt3ch, line 10: We haven't encountered that issue, yet, on the bootstrap script.

inthecloud247, line 10: It's mentioned here in the context of chef recipes:  http://stackoverflow.com/questions/9576124/what-is-the-right-way-to-do-add-apt-repository-via-chef

and here:
http://stackoverflow.com/questions/8445893/add-apt-repository-ppaferramroberto-java-how-to-add-in-a-bash-script

akoumjian:
This is very odd. Is it a known bug? I use add-apt-repository all the time in scripts and have never had a problem with '-y'.

inthecloud247:
It may be that it only appears the first time that add-apt-repository is run? I don't really know, except that I was demoing salt-cloud to someone today and got that 'Press [enter] to continue', and it was pretty annoying. I'm using ubuntu 12.04-1 LTS, which may also have some quirky behavior. I'd rather just see that it works :-)

akoumjian:
I use 12.04-1 pretty much exclusively. I am unable to reproduce this problem by running the script manually on ec2 or vagrant 12.04-1 VMs. The only thing I can think is that perhaps it is the way it is being executed?

Since this is one of the legacy scripts I would be hesitate to change their mechanics on the odd chance that someone is using < 10.04. That said I've tested this PR and it works fine in 12.04, as expected.

ghost:
add-apt-repository is in a a different package on ubuntu quantal and above.

http://packages.ubuntu.com/search?suite=quantal&arch=any&mode=exactfilename&searchon=contents&keywords=apt-add-repository

s0undt3ch:
@seanchannel Thanks for the info. Bootstrap script updated :)

inthecloud247:
wow good catch! can't believe Canonical changed it. (I guess that's the difference between LTS releases and regular ubuntu ones?) To avoid this in the future, wouldn't it make sense to have different ubuntu deploy scripts for different ubuntu versions? we have a rhel5 and a rhel6 script for example... maybe have ubuntu-12.04.sh , etc.? (of course this is for people who don't end up using the salt-bootstrap script, which I guess will need to handle all of these issues in one script.)

ghost:
add-apt-repository is just a convenience script for the wget/etc operation and creates an additional dep. that is why it was switched in the first place. the user action can be resolved without that.

inthecloud247:
yeah understood there... I prefer using packages where possible to avoid needing to worry about maintaining the script in case things change. It's definitely unusual to have the location of the packages switch... that kinda undermines the whole point of having the convenience script in a package in the first place.

PR #270, opened 2013-01-23
longname for update_salt_bootstrap function changed to update_bootstrap

inthecloud247:


saltcloud/utils/parsers.py
```diff
@@ -199,7 +199,7 @@ def _mixin_setup(self):
             help='Build all of the specified virtual machines in parallel'
         )
         group.add_option(
-            '-u', '--usb',
+            '-u', '--update_bootstrap',
```
techhat: This would need to be --update-bootstrap, not --update_bootstrap.

inthecloud247:
k this should be good now.

techhat:
You're going to want to start using feature branches in git, to keep those merge commits out.

inthecloud247:
yeah just noticed that... :-) thanks!

PR #272, opened 2013-01-23
More multi-region and tags

mafrosis:
Fixed two things:
- support multi region in all the new AWS functions (start,stop,set_tag,get_tag,del_tag)
- fix a bug around printing tags are renaming a node with `set_tags Name=something`

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -312,7 +312,8 @@ def stop(name):
     '''
     Stop a node
     '''
-    conn = get_conn()
+    location = get_location()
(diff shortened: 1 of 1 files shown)


mafrosis:
I was also wondering about renaming instances in this way.. We would need to reflect this in Salt too, since you would need to use the old name there. `salt-key` would return the wrong name too.

I don't really know how Salt maps names to IP/fqdn for minions on the inside, so I have no idea how involved this change might be.

techhat:
@mafrosis, I'll file another issue that addresses renaming.

PR #274, opened 2013-01-24
Add rename to aws

techhat:
Resolves issue #273.

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -406,3 +406,20 @@ def del_tags(name, kwargs):
         log.error('Failed to delete tags from {0}'.format(name))
         log.error(exc)
 
+
+def rename(name, kwargs):
+    '''
+    Properly rename a node
+    '''
+    location = get_location()
+    conn = get_conn(location=location)
+    node = get_node(conn, name)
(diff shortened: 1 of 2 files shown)


mafrosis:
How come it's `kwargs['newname']` and not just `newname`?

Please tell me to sssh if I'm just second guessing something that should be obvious ;0

techhat:
Imagine the following command:

```
salt-cloud -a rename name1 name2
```

In this instance, it would attempt to run the rename action against a machine named name1, and then a machine named name2... and it would fail, because it wouldn't know what the new name was for either.

With today's additions, if --action detects an argument with '=' in it, it turns it into a kwarg. And if it detects kwargs, it passes them through to the resulting action.

mafrosis:
Yeah.. I missed the implications of 8301e09 earlier. I'll just be quiet now.

techhat:
No worries, it's not obvious, and I bet you won't be the last person to question it.

mafrosis:
It's probably worth adding a usage line in the comment to that method. And maybe we should add a check to prevent people spinning up new nodes with an equals sign in their name. Like I just did.

techhat:
Good call. Let me add that to this pull req.

techhat:
Done.

PR #277, opened 2013-01-24
universal repo setup

ghost:
These comments are included in the commit:
# add-apt-repository requires an additional dep and is in different packages
# on different systems. Although seemingly ubiquitous it is not a standard,
# and is only a convenience script intended to accomplish the below two steps
# doing it this way is universal across all debian and ubuntu systems.

### DIFF
--- a/saltcloud/deploy/Ubuntu-git.sh
+++ b/saltcloud/deploy/Ubuntu-git.sh
@@ -5,6 +5,10 @@ echo '{{ vm['priv_key'] }}' > /etc/salt/pki/minion.pem
 echo '{{ vm['pub_key'] }}' > /etc/salt/pki/minion.pub
 echo "{{ minion }}" > /etc/salt/minion
 
+# add-apt-repository requires an additional dep and is in different packages
+# on different systems. Although seemingly ubiquitous it is not a standard,
+# and is only a convenience script intended to accomplish the below two steps 
+# doing it this way is universal across all debian and ubuntu systems.
 echo deb http://ppa.launchpad.net/saltstack/salt/ubuntu `lsb_release -sc` main | tee /etc/apt/sources.list.d/saltstack.list
 wget -q -O- "http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0x4759FA960E27C0A6" | apt-key add -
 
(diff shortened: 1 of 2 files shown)


akoumjian:
So... what about the claim that this code was hanging on 12.04.1? I wasn't able to reproduce, but that was why it changed.

inthecloud247:
yeah I don't know why this was reverted... if you check the history, this has now gone back and forth a few times.

inthecloud247:
it'd be more compatible to just have a check for the release and use the proper deb package.

akoumjian:
Considering that this is only supposed to be left here as legacy anyways, it seems like it should be keep the way it was before the first change. New users should be using bootstrap-script.

inthecloud247:
well if it's only legacy then we shouldn't care about it breaking for ubuntu > 12.04 ?

bootstrap-script was broken the last I checked a few days ago...

ghost:
please see gigantic comments in PR: https://github.com/saltstack/salt-cloud/issues/277

techhat:
I'm going to agree with @akoumjian here... nobody should be using these scripts anymore; they are now intended for academic purposes only. Let's stop going back and forth on this, and focus on getting salt-bootstrap to resolve any remaining issues.

ghost:
add-apt-repository's main purpose in life is to present the ubuntu user with nicely formatted prompts. It's a Canonical special and I highly recommend a more standard approach in salt-bootstrap as well.

inthecloud247:
@techhat If that's the case, then just delete them all please. It's really confusing to have old unmaintained scripts here.

If they remain in the project, clearly they'll need to be maintained to ensure that they continue to work with the newer versions of salt-cloud.

inthecloud247:
and @seanchannel, the point of the ubuntu.sh script is to install salt-cloud on ubuntu systems. debian systems should use the debian.sh script right? ubuntu is its own thing, and it's appropriate to treat it differently.

techhat:
@ydavid365, I will not remove functionality just because it confuses you, whether or not I think others should be using it. A number of users will not switch to salt-bootstrap immediately, and pulling out the rug underneath them like that would be unacceptable.

@seanchannel is a core Salt dev and our Ubuntu maintainer. As far as I am concerned, his word on Ubuntu is gospel until otherwise proven. Neither he, nor @akoumjian seem to be able to reproduce your issue. If you can give exact steps that @seanchannel can use to reliably and predictably reproduce the situation that is causing you stress, I have no doubt that he will do what it takes to resolve the issue, even if it means reversing his own changes. "Sometimes it doesn't do what I expect" isn't helpful. As it stands, I am content to leave these files as they are.

inthecloud247:
Here's the cloud.profiles entry on aws that was failing for me. Let me know if it works for you?

``` python
small_region:
  provider: aws
  image: ami-7e2da54e
  location: us-west-2
  availability_zone: us-west-2c
  size: Small Instance
  script: Ubuntu
```

PR #278, opened 2013-01-26
Termination protection enhancement

mafrosis:
I'm not sure if we want to merge this just yet, but I opened this pull req for discussion.

Added support for toggling termination protection using the botocore lib which is now [0.5.1 in Pypi](http://pypi.python.org/pypi/botocore/0.5.1).

Also wrapped the core libcloudfuncs.destroy() method to gracefully handle the exception raised in libcloud when termination protection is switched on for a node.

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -477,3 +477,69 @@ def rename(name, kwargs):
             )
         )
         log.error(exc)
+
+
+def enable_term_protect(name):
+    '''
+    Enable termination protection on a node
+    '''
+    _toggle_term_protect(name, True)
+
+
+def disable_term_protect(name):
+    '''
+    Disable termination protection on a node
+    '''
+    _toggle_term_protect(name, False)
+
+
+def _toggle_term_protect(name, enabled):
+    '''
+    Toggle termination protection on a node
+    '''
+    try:
+        import botocore.session
+    except ImportError:
+        log.error('You must install module botocore')
+        return
+
+    # region is required for all boto queries
+    region = get_location(None)
+
+    # init botocore
+    session = botocore.session.get_session()
+    session.set_credentials(
+        access_key=__opts__['AWS.id'],
+        secret_key=__opts__['AWS.key'],
+        )
+
+    service = session.get_service('ec2')
+    endpoint = service.get_endpoint(region)
+
+    # get the instance-id for the supplied node name
+    client = salt.client.LocalClient()
+    ret = client.cmd(name, 'grains.item', ['instanceId'])
+
+    if ret[name] == "":
+        log.error("Couldn't get instance_id for {0}".format(name))
+        return
+    else:
+        instance_id = ret[name]
+
+    params = {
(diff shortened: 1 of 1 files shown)


techhat:
I'm pretty excited about the idea of using botocore. I'm not sure about the late import on this. Maybe we can just doc that botocore is only needed to be installed for advanced actions? Perhaps @thatch45 can chime in when he gets a chance.

mafrosis:
There's a number of things in there that I'm not too sure about - the late import being one of them.

I was thinking that some advanced features would rely on botocore - then when libcloud catches up, we can port the code to use that. We don't want to add a dependency on botocore (since it's only for AWS) but the people who use AWS will almost certainly want the extra features provided by boto...

techhat:
@mafrosis, you'll want to take a look at this again... it looks like whatever changes have been applied since we first started looking at it have made it unmergeable (is that a word?).

mafrosis:
Yeah I'll need to rebase the code periodically since there's a lot happening in the aws module right now.

The question is still around how we're going to work with botocore.. If people are happy to `pip install` in the short term, then this can be merged..

techhat:
Is there anything that libcloud does that botocore doesn't do? I'm not opposed to switching the AWS module entirely over to botocore, if it's a better solution.

mafrosis:
I'm certain everything _can_ be done by boto (or will be in the near future) - however porting the AWS will be a bit of work. The salt-cloud config is tightly coupled to the params passed to libcloud and that will all need resolving.

Start small with just the functions missing from libcloud?

techhat:
Sounds like a plan. But maybe before we do that, we should fill in the rest of what libcloud has that we're not using, so we know what needs to be filled in.

mafrosis:
How about a policy of "use libcloud if possible, else use botocore". At least until we have botocore in apt/yum and we can port the whole module.

techhat:
Sold.

techhat:
This is good, and it looks like it can be merged again... but it's missing the new "delete_sshkeys" option that I added to libcloudfuncs yesterday in #324. Would you mind adding that before I merge this?

mafrosis:
Updated to latest develop HEAD

techhat:
Nice to finally be able to merge this!

mafrosis:
Agreed! Since 0.13.0 has just dropped, is it time to roll another salt-cloud release? (Or has that happened when I wasn't watching?)

techhat:
The time has come (especially with the config breakage between salt 0.13.0 and salt-cloud <= 0.8.4). I'll be cutting a release this week.

mafrosis:
I've just been reviewing the code we just merged, and I have a couple of comments/changes. Sorry I didn't do this earlier!
- All EC2 calls in botocore use the node's "instanceId". This is only available through salt when the "ec2_info" grain is available.. Which complicates things greatly.
- We still have the late import. This should be changed.
- I added usage examples to the new methods.

techhat:
Well, let's get another pull req open, and we'll get 'em merged. Thanks for all the work, @mafrosis!

PR #279, opened 2013-01-26
add CLI flag to assume yes to confirmation prompts

mafrosis:
..

### DIFF
--- a/saltcloud/utils/parsers.py
+++ b/saltcloud/utils/parsers.py
@@ -204,6 +204,12 @@ class ExecutionOptionsMixIn(object):
             action='store_true',
             help='Update salt-bootstrap to the latest develop version on GitHub'
         )
+        group.add_option(
+            '-y', '--assume-yes',
+            default=False,
+            action='store_true',
+            help='Default yes in answer to all confirmation questions'
+        )
         self.add_option_group(group)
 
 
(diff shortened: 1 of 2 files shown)


saltcloud/utils/parsers.py
```diff
@@ -204,6 +204,13 @@ def _mixin_setup(self):
             action='store_true',
             help='Update salt-bootstrap to the latest develop version on GitHub'
         )
+        group.add_option(
+            '-y', '--assume-yes',
+            default=False,
+            dest='assume_yes',
```
techhat: Check out #266, and take note of the absence of dest.

mafrosis: I was imitating the "--assume-yes" option from aptitude. The dash means it won't directly work as a variable name - so I used dest, as above on the "--no-deploy" option.

s0undt3ch: Long `OptionParser` options have their `dest` set automatic to something like:

``` python
'--long-option'.lstrip('--').replace('-', '_')
```

If not explicitly set.

PS: Above is untested code, but you get the idea.

mafrosis: Thanks for the pointer Pedro - I didn't know that.

mafrosis:
Pull req updated to remove unnecessary config param

techhat:
Excellent.

saltcloud/utils/parsers.py
s0undt3ch: If I can help :smiley:

PR #281, opened 2013-01-26
install conf as examples

ghost:
one-line change to install conf/ at /usr/share/doc/salt-cloud/examples


ghost:
the aws examples are all official images for Debian, Ubuntu, and Fedora (so far)

techhat:
I really like that all of these are tagged with the region that they go in (assuming us-west-1 is the default region for them), but there's an awful lot of them, and they're all biased towards US users.

What if we were to rename cloud.profiles to cloud.profiles.us-west-1? We should also put a disclaimer in there so nobody thinks these are paid advertisements. And if we're going to keep these in here, we should check on them from time to time to make sure they're still valid (and datestamp each one in the comments as to when they were last verified).

What do you think? Maybe @mafrosis could fill in some Australian examples?

ghost:
I agree it's biased.  Note that this is not going directly into cloud.profiles, but a new file presently named aws.profiles.  Perhaps rename this as you say and list all the same images for other regions in another such, and so on. I'm trying to cover the basic set of distros with officially sanctioned amis, and plan to do this for all providers as much as possible.

I don't want to create a big maintenance burden here. I see this as useful to myself and that's mainly what's driving it more than 'what would be a good example'

PR #282, opened 2013-01-27
Don't require make_master config option in cloud.profile

mafrosis:
Diff should be self-explanatory - an exception is thrown without this config option present when creating a new node.


techhat:
Oh geez, how embarrassing. Thanks for cleaning that up for me.

PR #293, opened 2013-01-30
Include cli options into loaded config dictionary.

s0undt3ch:
The merge with cli mix in, only merges option keys which exist in both places, cli options and loaded config. Since neither `profile` not `parallel` existed in any of the default config dictionary, not the loaded config files, that info was not getting included.

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -11,8 +11,8 @@ Primary interfaces for the salt-cloud system
 # the VM data will be in opts['vm']
 
 # Import python libs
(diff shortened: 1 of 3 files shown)


s0undt3ch:
@techhat is this enough?

s0undt3ch:
This fix was not the issue.

s0undt3ch:
Ok, see if this solves our issues.

techhat:
Looks much better. Thanks, @s0undt3ch!

PR #296, opened 2013-01-31
Also accept mode 0400

techhat:
Resolves issue #290.

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -76,9 +76,9 @@ def __virtual__():
     keymode = str(
         oct(stat.S_IMODE(os.stat(__opts__['AWS.private_key']).st_mode))
     )
-    if keymode != '0600':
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Consider the following example script:

``` python
import timeit

def test_in():
    entries = [
        '0300', '0400', '0500', '0600'
    ]
    for entry in entries:
        if entry in ('0400', '0600'):
            pass

def test_double_if():
    entries = [
        '0300', '0400', '0500', '0600'
    ]
    for entry in entries:
        if entry == '0400' or entry == '0600':
            pass


print 'Test In'
print(timeit.timeit("test_in()", setup="from __main__ import test_in"))
print 'Test Double If'
print(timeit.timeit("test_double_if()", setup="from __main__ import test_double_if"))
```

And that scripts output:

```
Test In
0.894567966461
Test Double If
1.0446879863
```

PR #299, opened 2013-01-31
Some code linting love

s0undt3ch:
- Pylint fixes
- PEP-8
- Code formatting
- Unused imports
- etc....

Initially pylint reported an initial score of `7.52/10`, it's now `8.23/10`.


techhat:
This is intense, but in going over the commits with @s0undt3ch, I think we're good.

PR #301, opened 2013-01-31
Fix #290 by allowing 0400 perms on a key

techhat:
Thanks to @s0undt3ch for pointing out the cesspool of badness in #296.


s0undt3ch:
Heh, i just liked how it look better. The speed test was just to make sure I wasn't suggesting a bad thing ;)

PR #302, opened 2013-01-31
Allow user to keep tmp files without code diving

techhat:
This will save time when troubleshooting salt-bootstrap. @s0undt3ch may especially appreciate this.


s0undt3ch:
Yep! Might prove very helpful...

PR #305, opened 2013-01-31
Allow script_args to be passed to deploy.sh

techhat:
Resolves #218. Sample configuration might look like:

``` yaml
aws-amazon:
    provider: aws
    image: ami-1624987f
    size: Micro Instance
    ssh_username: ec2-user
    script: bootstrap-salt-minion
    script_args: -c /tmp/
```

This will help save some users from having to write wrapper scripts, just to pass options along to salt-bootstrap.

Tagging @s0undt3ch, @akoumjian.

### DIFF
--- a/saltcloud/clouds/aws.py
+++ b/saltcloud/clouds/aws.py
@@ -295,6 +295,7 @@ def create(vm_):
             'minion_pem': vm_['priv_key'],
             'minion_pub': vm_['pub_key'],
             'keep_tmp': __opts__['keep_tmp'],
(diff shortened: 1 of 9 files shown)


s0undt3ch:
These are direct script arguments right? Not piped arguments?

techhat:
Anything you want to just attach to the end of the command that gets run. Some of my testing was performed with an argument of:

```
'| cat'
```

s0undt3ch:
Ok, the possible issue was piping to sh itself.

PR #313, opened 2013-02-06
Sudo is being prepended later if needed.

ConsoleCatzirl:
If sudo is needed, it can be set in the cloud profile.  If it is
set there, then we end up with 'sudo sudo /tmp/deploy.sh', which
just looks ugly.


techhat:
Good catch, thanks duder!

PR #314, opened 2013-02-06
Fallback to using a password if there is no pubkey given.

ConsoleCatzirl:
Without this, if no public key file is specified, you get an easy-to-miss error during the deploy and no minion.


techhat:
I would love to use this, but I can't gotten it to work yet.

techhat:
@jesusaurus, I'm afraid I've pulled the underlying code out from under you. The good news is, the code that I added still accomplishes what you were looking for, and was inspired entirely by this pull req.

inthecloud247:
cue bad big lebowsky-inspired quote:

u don't mess with the @jesusaurus :-)

great idea here guys. cool.

PR #316, opened 2013-02-06
Replace Paramiko with sshpass

techhat:
Using Paramiko has caused us nothing but pain, and we were only using it because we didn't have a way to pass a password to the ssh command. The sshpass package is tiny and widely available. The tradeoff here in terms of performance is that we're giving up the overhead of a massive SSH library that we only used a fraction of, for the overhead of shelling out more. I'm okay with that.

@herlo, @seanchannel and @cedwards will want to be aware that we have removed paramiko as a dependency, and added sshpass. I've been able to find it in epel, packages.debian.org, packages.ubuntu,org and even in the Arch community repo.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,9 +22,9 @@ Depends: ${python:Depends},
          python,
          python-libcloud,
          python-boto,
-         python-paramiko,
          python-yaml,
-         salt-common
+         salt-common,
+         sshpass
 Description: public cloud VM management system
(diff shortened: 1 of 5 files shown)


ghost:
I uploaded salt-cloud 0.8.4-138-gfc3e5e0 to the daily PPA, FWIW. HTH! :)

```
add-apt-repository ppa:saltstack/salt-daily
```

techhat:
Huzzah!

PR #321, opened 2013-02-08
update script name

ghost:
yar! ya best be updatin' this one, too! "bootstrap-salt-minion" was recently renamed to several variants.

### DIFF
--- a/saltcloud/config.py
+++ b/saltcloud/config.py
@@ -15,7 +15,7 @@ CLOUD_CONFIG_DEFAULTS = {
     'ssh_auth': '',
     'keysize': 4096,
     'os': '',
-    'script': 'bootstrap-salt-minion',
+    'script': 'wget-bootstrap',
     'start_action': None,
     # Logging defaults
     'log_file': '/var/log/salt/cloud',


techhat:
I knew they were planning on renaming, though I haven't looked yet. However, I don't think wget is a good default, since it can rarely be expected to be there on a new system. Plus, official salt-cloud releases will contain a full copy of the actual bootstrap script.

ghost:
As it stands now it's broken; no deploy script gets called at all due to the default being one that doesn't exist. perhaps the default in config.py needs to point at something that exists and works now, but can be changed later, like "default.sh", "bootstrap-generic", etc.

techhat:
To be fair, it also didn't work straight out of git before the name change either. We only bundle the script with the release tarball.

The problem with making anything else the default is that they rely on other components that can't be expected to be on the system already. Neither curl nor wget is standard across the board (and are almost never installed by default), and even Python can't be expected to be installed in FreeBSD, which we also have active users in.

So as long as you install from the release tarball (via PyPi or one of the community packages), you'll be fine, because the bootstrap script is there, and by design needs no other components. We could, however, put together a doc that discusses running an instance from a git checkout.

ghost:
So there is an as-yet-unwritten script that is not yet included?  Something that exists has to be the default. I understand if there is another script pending, but it looks like bootstrap has been added already. This seems intentionally left broken for reasons I do not fathom.

techhat:
But... something that exists in the release tarball is the default. It's not pending or anything, we just include the latest stable of salt-bootstrap when we cut a release.

ghost:
It would be helpful to have that "latest stable" version committed in salt-cloud. Anyone using git will need to know how to deal with this.

inthecloud247:
What about including salt-bootstrap as a git submodule?

PR #323, opened 2013-02-11
Handle the rename of the bootstrap script

mafrosis:
The recent rename in salt-bootstrap caused me a few issues, and then it seemed pertinent to update salt-cloud to match.

I haven't done anything to handle salt-cloud config containing "script: bootstrap-salt-minion", as yet. Really that should be removed from people's config since it's now the default..

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -49,17 +49,13 @@ class SaltCloud(parsers.SaltCloudParser):
 
         if self.options.update_bootstrap:
             import urllib
-            branch = 'develop'
-            url = (
-                'https://raw.github.com/saltstack/salt-bootstrap/{0}/'
-                'bootstrap-salt-minion.sh'.format(branch)
-            )
+            url = 'http://bootstrap.saltstack.org'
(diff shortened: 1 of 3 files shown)


thatch45:
Thanks @mafrosis!

saltcloud/cli.py
```diff
@@ -49,17 +49,13 @@ def run(self):
 
         if self.options.update_bootstrap:
             import urllib
-            branch = 'develop'
-            url = (
-                'https://raw.github.com/saltstack/salt-bootstrap/{0}/'
-                'bootstrap-salt-minion.sh'.format(branch)
-            )
+            url = 'http://bootstrap.saltstack.org'
```
techhat, line 52: @s0undt3ch, will this give us the latest version in git, or the latest stable version?

s0undt3ch, line 52: It still gives us latest "unstable". We still need to get the FreeBSD repository issue solved and re-run the script in all supported distributions. After that, a master branch will be created, and we should update `http://bootstrap.saltstack.org` to point to the stable branch.

PR #324, opened 2013-02-11
Allow salt-cloud to remove ssh keys upon destroy()

techhat:
In order to use this, add this to your main salt-cloud config file:

``` yaml
delete_sshkeys: True
```


techhat:
Resolves #317. Tagging @jesusaurus.

PR #330, opened 2013-02-13
new ubuntu amis

ghost:
I might just push this sort of thing if @techhat didn't mind.


techhat:
What are the differences between these AMIs and the old ones?

ghost:
These are the official ones from canonical. They just released these updated amis recently. The old/existing amis come up with a MOTD about the new images. These all originate from http://cloud-images.ubuntu.com/releases/

techhat:
K, I figured it was something like that, but wanted to make sure.

PR #332, opened 2013-02-13
fix URL appearances

ghost:
they was hidden!


techhat:
I'm embarrassed that this wasn't caught before now. Thanks for getting this.

PR #336, opened 2013-02-13
Let users know that a map is being applied.

s0undt3ch:


s0undt3ch:
Also "Avoid `KeyError`'s when there ain't a matching provider for a mapped profile."

PR #344, opened 2013-02-18
info default

inthecloud247:


techhat:
@ydavid365, this was also caught in my refactoring of the config file recently, which was a little more expansive. There was a lot of formatting that needed fixing, and some other examples needed to be added or removed.

PR #345, opened 2013-02-18
2013 year

inthecloud247:


techhat:
Hi @ydavid365. Apologies for not catching this earlier, I was pretty busy at the beginning of this week, and not paying as much attention to pull reqs as I should have been. It looks like some of the cleanup I did earlier already took care of the year in the license file. I guess I could have saved myself some work had I been paying attention. Also, the man pages are automatically generated at release time, so any changes will be overwritten at that time.

PR #356, opened 2013-02-21
Add cloud-specific functions (separate from actions)

techhat:
This changes the way that -a/--action is handled internally, and keeps -f/--function functions from being called by -a, and keeps actions from being called by -f.

Sample usage for this style looks like:

```
salt-cloud -f ec2 show_image image=ami-fd20ad94
```

If such a function were to be called with --action, an instance name would be required to make the call, but not used.

### DIFF
--- a/doc/topics/function.rst
+++ b/doc/topics/function.rst
@@ -1,0 +1,13 @@
+===============
+Cloud Functions
+===============
+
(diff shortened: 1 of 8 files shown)


techhat:
This concept around this pull req is currently being evaluated by the core devs, so please do not merge it without talking to @thatch45 or @techhat first.

PR #357, opened 2013-02-22
The `--function` parser option now expects two arguments.

s0undt3ch:
- Additionally we fail if the desired provider does not implement the function we want.


techhat:
The two arguments being "function name" and "provider name"?

PR #361, opened 2013-02-22
Cleaning up old debug stuff from ec2 module

techhat:


UtahDave:
Brilliant work, @techhat

PR #366, opened 2013-02-22
Resolve stacktrace from destroying protected instances

techhat:
Resolves #363.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -641,6 +641,14 @@ def destroy(name, call=None):
     '''
     nodes = list_nodes_full()
     instance_id = nodes[name]['instanceId']
+    protected = show_term_protect(instance_id=instance_id,
+                             call='action',
+                             quiet=True)
+    if protected == 'true':
+        print('This instance has been protected from being destroyed. Use the '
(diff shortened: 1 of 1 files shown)


saltcloud/clouds/ec2.py
```diff
@@ -754,10 +762,13 @@ def show_term_protect(name, instance_id=None, call=None):
...
-    if disable_protect == 'true':
-        print('Termination Protection is enabled for {0}'.format(name))
-    else:
-        print('Termination Protection is disabled for {0}'.format(name))
+    if quiet is False:
+        if disable_protect == 'true':
+            print('Termination Protection is enabled for {0}'.format(name))
+        else:
+            print('Termination Protection is disabled for {0}'.format(name))
+
```
s0undt3ch, line 772: :smile: like?

``` python
log.log(
    logging.DEBUG if quiet is True else logging.WARNING,
    'Termination Protection is {0} for {1}'.format(
        disable_protect == 'true' and 'enabled' or 'disabled',
        name
    )
)
```

techhat, line 772: Addressed.

s0undt3ch:
Seems pretty good to me.

PR #368, opened 2013-02-23
Add direct avail_sizes function to ec2

techhat:
It should be noted that this also changes the format in which the data is returned from avail_sizes. However, it is much more accurate and informative than what we were returning before from libcloud.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -104,15 +104,13 @@ def __virtual__():
             )
         )
 
-    global avail_images, avail_sizes, script, destroy
-    global list_nodes_select
+    global avail_images, script, list_nodes_select
 
     # open a connection in a specific region
     conn = get_conn(**{'location': get_location()})
 
     # Init the libcloud functions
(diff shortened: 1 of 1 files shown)


techhat:
@thatch45, could I get your opinion before we merge this?

s0undt3ch:
The list cannot be computed at runtime? Using the direct API? You'd have to keep updating that list(dict) every time it changed...

techhat:
This is exactly how libcloud does it:

https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/ec2.py

But I was very careful not to copy their code. Most of my research was based on the information in the URL I put in the docstrings, the rest was in trying to keep the return data look vaguely like how our AWS driver was already presenting it, but more complete.

s0undt3ch:
If there's not way to get that information using EC2 api, then hardcoded is the way to go. (/me thinks)

techhat:
I'm going to go ahead and merge this. If it ends up being the wrong direction, we can back it out.

PR #370, opened 2013-02-23
Sync after install

mafrosis:
This is a small feature that's been useful for me on AWS - it will get salt-cloud to synchronise any custom grains/modules/states over to a minion right after create.

It came about because I needed the ec2_info grain available right after a minion was deployed - before waiting for highstate to run.

### DIFF
--- a/doc/topics/misc.rst
+++ b/doc/topics/misc.rst
@@ -22,3 +22,26 @@ the following line needs to be added to the main cloud configuration file:
 
     delete_sshkeys: True
 
+Sync After Install
+==================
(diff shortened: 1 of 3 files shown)


techhat:
Thanks @mafrosis, this looks good. @ydavid365, does this address what you were getting at in #348?

inthecloud247:
@techhat I don't think it'll help with #348 since the grains being referred to here are custom grains modules, and in #348, I'm referring to grains set specifically in the /etc/salt/grains file.

PR #373, opened 2013-02-25
Allow EC2 driver to use salt outputter

techhat:
This will cause all output from EC2 to be returned through the salt outputter, and enable all other cloud providers to do the same thing.


s0undt3ch:
Actually use salt's outputters, make sense...

PR #377, opened 2013-02-25
Adventures in the Python regex module

mafrosis:
The diff says it all


techhat:
This is better, thanks, duder!

PR #381, opened 2013-02-25
CLI mods

mafrosis:
Previously we use exits all over to break the flow in the cli module. I believe this patch is a better solution!

Comments please.

Changes:
- correct exit codes on error
- refactored to use and if-else block, so we only exit on success in one place
- changed query option to be exclusive of other options; the same as others in the list (this the only functionality change; look at bbbd10c)

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -96,6 +96,7 @@ class SaltCloud(parsers.SaltCloudParser):
                             exc
                         )
                     )
+                    self.exit(1)
             else:
                 try:
                     query_map = mapper.map_providers(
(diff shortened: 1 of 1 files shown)


techhat:
This is pretty intense. Let me take my time looking through this before I merge it.

mafrosis:
No worries. I tried to break down the commits as far as was sensible, so it'll be easier to digest that way.

techhat:
Actually, let me merge @s0undt3ch's first (#380), which is likely to cause a merge conflict. Keep posted.

mafrosis:
Hehe yeah I saw that coming too. I'll rebase over here when you're ready.

techhat:
Looks good, let's go with it.

PR #382, opened 2013-02-26
Outputter changes

mafrosis:
- Now using the output from create to indicate success for the sync_after_install option 
- Output some results create/destroy with a map
- Use the salt command line options for format our output!


techhat:
PEP8 suggestion: "is not False" instead of "!= False".

mafrosis:
@techhat bah I noticed s0undt3ch modified another of my if statements elsewhere with that very thing

Python isn't my first language ;)

PR #121, opened 2012-11-08
Ability to deploy EBS volumes on ec2

maratsh:
Latest libcloud has support EBS management.
This commits is  just draft for volumes attaching support.

example:

```
test-micro:
    - test-temp-s02:
        volumes:
            - { size: 1, device: /dev/sdf}
            - { size: 1, device: /dev/sdh}
```

PR #383, opened 2013-02-26
Add volume management to ec2

techhat:
The only one of these that requires a VM name or instance_id is attach_volume(), which also requires a volume_id. create_volume() requires an availability zone (zone), and either a size in GiB (default 10) or a snapshot id (snapshot) to make a copy of. show_volume, detach_volume and delete_volume only require a volume_id.


techhat:
This is related to #121.

PR #385, opened 2013-02-26
Option to disable output from SSH deploy script

mafrosis:
I changed from subprocess.call to using the communicate function. This should ensure that we consume the output from the SSH connection, but we don't actually need to print it on stdout.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -629,6 +629,9 @@ def create(vm_=None, call=None):
             'minion_pub': vm_['pub_key'],
             'keep_tmp': __opts__['keep_tmp'],
         }
+        if 'display_ssh_output' in __opts__:
+            deploy_kwargs['display_ssh_output'] = __opts__['display_ssh_output']
+
         deploy_kwargs['minion_conf'] = saltcloud.utils.minion_conf_string(
             __opts__, vm_
         )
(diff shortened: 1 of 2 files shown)


techhat:
Testing this on a local branch, I was still able to see the SSH output of the deploy script, using the following profile:

``` yaml
ec2-ubuntu:
    provider: ec2
    image: ami-3d4ff254
    size: Micro Instance
    script: bootstrap-salt
    ssh_username: ubuntu
    delvol_on_destroy: True
    display_ssh_output: False
```

Did I misunderstand the usage?

techhat:
My apologies, I was confusing `__opts__` with `vm_`. This is working fine, when I move the option into the main cloud config, which is what makes sense with trying to spin up instances in parallel.

I'm going to go ahead and merge this, but could you please add usage documentation to the doc/ directory?

mafrosis:
Nice! Honestly, I didn't know if this would be accepted.. Since it now has, I'll write the docs entry.

techhat:
You're not the first person to mention how noisy the output is, so I'm going to assume it's a feature other people may want. And I'm not going to go all Gnomish on you and start pulling out feature you use, just because they don't suit my own needs. So yeah, I'm okay with accepting it. ^_^

mafrosis:
Well nobody likes a load of uncontrollable noise being dumped on stdout right? I was thinking that maybe log level debug should auto display_ssh_output to true. What do you think?

techhat:
I always watch the output from SSH, but uncontrollable output is another thing. And I would be hesitant to auto-set that option, since I always use debug mode, and may have a need to turn off SSH output.

PR #387, opened 2013-02-26
Try to properly handle `get_node` failures. Fixes #386.

s0undt3ch:


saltcloud/clouds/libcloud_aws.py
```diff
@@ -284,6 +298,9 @@ def create(vm_):
                     key_filename=__opts__['AWS.private_key']):
                 username = user
                 break
+    if data.public_ips:
```
techhat: Was this properly placed? It doesn't look right.

s0undt3ch: we need to break out of the while True loop. I'll do some more testing when I get home.

techhat: That's what I figured, but that's also not where it is.

s0undt3ch:
Should be good now ;)

PR #391, opened 2013-02-26
add boostrap submodule

ghost:
If you like..

this adds a reference to the stable branch salt-bootstrap in `saltcloud/deplo/bootstrap`.  On a fresh clone use `git clone --recursive` to also obtain bootstrap. On an existing clone you should be able to run `git submodule init; git submodule update` to bring this in.

Note that submodules point to a particular commit, so when salt-bootstrap has a new stable version available, the submodule reference needs to be updated with `git submodule update` and committed to refer to the new version. 

It should not be necessary to update or deal with submodules otherwise.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "saltcloud/deploy/bootstrap"]
+	path = saltcloud/deploy/bootstrap
+	url = https://github.com/saltstack/salt-bootstrap.git


inthecloud247:
awesome @seanchannel ! this should be great. Can you also update the docs with those instructions? Definitely seems to be a good use of submodules.

mafrosis:
You can also simplify that second command down to `git submodule update --init`

PR #396, opened 2013-02-28
A couple of small fixes

mafrosis:
- added docs entry fro display_ssh_output
- write map errors on STDERR


techhat:
Good call on both.

PR #397, opened 2013-02-28
Doc changes

ahammond:
I threw together some doc updates based on my experience getting `salt-cloud` working tonight. Big thanks to redbeard2 in #salt for all the help!

### DIFF
--- a/doc/topics/rackspace.rst
+++ b/doc/topics/rackspace.rst
@@ -1,0 +1,82 @@
+==============================
+Getting Started With RackSpace
+==============================
+
+RackSpace is the other major public cloud platform and is one of the core
+platforms that Salt Cloud has been build to support.
+
+Set up the cloud config at ``/etc/salt/cloud``:
+
+.. code-block:: yaml
+
+    # Set up an optional default cloud provider
+    provider: openstack
+
+    # Set the location of your salt master
+    minion:
+      master: saltmaster.example.com
+
+    # Set your RackSpace login data using the OpenStack plugin
+    OPENSTACK.identity_url: 'https://identity.api.rackspacecloud.com/v2.0/tokens'
+    OPENSTACK.compute_name: cloudServersOpenStack
+    OPENSTACK.protocol: ipv4
+
+    # You may need to change the compute_region
+    OPENSTACK.compute_region: DFW
+
+    # You will certainly need to change each of these
+    OPENSTACK.user: myname
+    OPENSTACK.tenant: 123456
+    OPENSTACK.apikey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+
+Under Hosting -> Cloud Servers, I see that my servers are provisioned in
+either DFW1 or DFW2, however my ``compute_node`` is set to DFW (no trailing
+digits).
+RackSpace has other regions, but I don't at the time of writing this know
+what the other regions are.
+If you do, please submit a pull-request to fix this.
+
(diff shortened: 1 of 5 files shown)


doc/topics/rackspace.rst
```diff
@@ -0,0 +1,82 @@
...
+the Salt Master.
+You can confirm this by running the following on the salt master.
+
+.. code-block:: bash
+
+    # salt 'foo' test.ping
+
+You should see your new node responding with a True.
+
+Optional Settings
```
techhat, line 76: If we don't have any optional settings at the moment, let's leave them out.

doc/topics/rackspace.rst
```diff
@@ -0,0 +1,82 @@
+==============================
+Getting Started With RackSpace
+==============================
+
+RackSpace is the other major public cloud platform and is one of the core
```
techhat: I wouldn't call it "the" other. I would rather not advertise any one or two cloud platforms above any others.

doc/topics/rackspace.rst
```diff
@@ -0,0 +1,82 @@
...
+
+    # You may need to change the compute_region
+    OPENSTACK.compute_region: DFW
+
+    # You will certainly need to change each of these
+    OPENSTACK.user: myname
+    OPENSTACK.tenant: 123456
+    OPENSTACK.apikey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+
+Under Hosting -> Cloud Servers, I see that my servers are provisioned in
```
techhat: First person should be avoided in technical documentation at all costs.

doc/topics/rackspace.rst
```diff
@@ -0,0 +1,82 @@
...
+Set up an initial profile at ``/etc/salt/cloud.profiles``:
+
+.. code-block:: yaml
+
+    openstack_512:
+        provider: openstack
+        size: 512MB Standard Instance
+        image: Ubuntu 12.04 LTS (Precise Pangolin)
+
+You can now instantiate a machine based on this profile with a salt
```
techhat: Second person should also be avoided. Maybe something like, "A machine based on this profile may now be instantiated with a salt command".

doc/topics/rackspace.rst
```diff
@@ -0,0 +1,82 @@
...
+command:
+
+.. code-block:: bash
+
+    # salt-cloud -p openstack_512 foo
+
+This will create a virtual machine at RackSpace with the name ``foo``.
+It took about 5 minutes to complete when I tried it.
```
techhat: Leave out exact times, so as not to raise expectations. I have seen this take anywhere from 2 to 15 minutes on Rackspace, and everyone's mileage will vary.

techhat:
Don't let all of my inline comments discourage you. This is good stuff that is sorely needed, but it does need a little work before it can be accepted.

ghost:
agreed @ahammond ! Please commit some updates per notes above & those commits will automatically appear here in the pull req.

doc/topics/rackspace.rst
ahammond, line 76: I think that positively saying that there are no optional settings is better. I'll update it that way.

doc/topics/rackspace.rst
ahammond: Reworded to clarify and avoid setting expectations.

techhat:
It's better, but there's still some first person in there. But I'm seeing some other first person as well, so I'm going to merge this, and then see if I can do some overall cleanup. Thanks for your help, @ahammond!

PR #404, opened 2013-03-04
Parallelization improvements

mafrosis:
Using this patch, we can now get the output from many parellized creates when running from a map file. The mods are all around the multiprocessing code.

Currently salt-cloud will start as many processes as there are new VMs to start - (this patch also behaves this way) - but I'm not sure if that's the best approach. Perhaps someone more knowledgable in this area can offer some wisdom..

Some benchmarks creating instances in EC2 from an existing m1.small VM in the same region:

```
servers secs
3       206
6       270
8       385
12      509
15      605
20      641
```

NB: this patch doesn't include anything around the `tty` flag to false when calling `deploy_script`. I need `tty` to be false to prevent my terminal getting mangled during the create.

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -472,6 +472,8 @@ class Map(Cloud):
         Execute the contents of the VM map
         '''
         output = {}
+        if self.opts['parallel']:
+            parallel_data = []
         # Generate the fingerprint of the master pubkey in
         #     order to mitigate man-in-the-middle attacks
         master_pub = self.opts['pki_dir'] + '/master.pub'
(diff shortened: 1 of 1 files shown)


inthecloud247:
Cool!

techhat:
@thatch45, multiprocessing is not my strong point, would you mind taking a look at this?

thatch45:
Looks good, that is how to use a pool, much more elegant than my original approach

mafrosis:
Cool beans.

@thatch45 any thoughts on setting the size of the Pool? Or is it acceptable to just create as many processes as there are VMs and let the OS work it out?

PR #408, opened 2013-03-05
Allow specifying the deploy value from the `vm` profile.

s0undt3ch:
Now, the `vm` deploy value is grabbed initially from the `vm` profile, and if not set, grab the value from `__opts__`.

I got the idea that this was supported from @cedwards presentation slides. If it was supported, that support got lost and this pull request fixes it.

Should we also check if the could provider has a "global" `deploy` setting too?

``` python
vm_.get('deploy', __opts__.get('<provider>.deploy', __opts__['deploy'])) is True
```


s0undt3ch:
Let me know if we should also implement the global provider deploy setting too.

PR #409, opened 2013-03-05
Allow setting `delvol_on_destroy` for **all** EC2 VMs. PEP-8.

s0undt3ch:


techhat:
Be aware that in 0.8.7, we're switching xml_to_dict over to salt.utils.xmlutil.

s0undt3ch:
Hmm, and the implications of that regarding these changes?

techhat:
Just that the _xml_to_dict function in here will disappear within the next month anyway.

s0undt3ch:
Ah :smile: it was just a formatting change.

PR #430, opened 2013-03-09
switch provider from aws to ec2

ghost:
maybe it's about time to do this... or just foil another one of my schemes. :)


techhat:
Let's hold off on this until we declare the driver to be stable.

PR #432, opened 2013-03-09
Address #403: Allow map to create master and associated minions

whodatninja:
When a make_master is present in a map, it's created first and used as the minion.master for any minions in the map lacking that attribute.

Unnecessary log output and inadvertently included bootstrap file removed.

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -545,7 +545,8 @@ class Map(Cloud):
         for miniondict in self.map[tvm['profile']]:
             if isinstance(miniondict, dict) and name in miniondict:
                 for option in option_types:
-                    tvm['map_{0}'.format(option)] = miniondict[name][option]
+                    if option in miniondict[name]:
+                        tvm['map_{0}'.format(option)] = miniondict[name][option]
         return tvm
 
 
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -484,25 +484,41 @@ class Map(Cloud):
         output = {}
         if self.opts['parallel']:
             parallel_data = []
+        master_name = None
+        master_host = None
+        try:
+            master_name, master_profile = ((name, profile) for name, profile in
+                dmap['create'].items() if 'make_master' in profile and
+                profile['make_master'] is True).next()
+            log.debug('Creating new master {0}'.format(master_name))
+            tvm = self.vm_options(master_name, master_profile, None, ['grains', 'volumes'])
+            out = self.create(tvm)
+            if 'deploy_kwargs' in out and 'host' in out['deploy_kwargs']:
+                master_host = out['deploy_kwargs']['host']
+                output[master_name] = out
+            else:
+                print('Host for new master {0} was not found, '
+                    'aborting map'.format(master_name))
+                sys.exit(1)
+        except StopIteration:
+            log.debug('No make_master found in map')
+
         # Generate the fingerprint of the master pubkey in
         #     order to mitigate man-in-the-middle attacks
         master_pub = self.opts['pki_dir'] + '/master.pub'
         master_finger = ''
         if os.path.isfile(master_pub) and hasattr(salt.utils, 'pem_finger'):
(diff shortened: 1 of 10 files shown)


techhat:
I've got bad news: there's a merge conflict in here. At the very least, I know version.py has a conflict, but you'll want to check the rest of it too.

mafrosis:
It's probably just me, but it would be nice if you could squash all those commits into just the one good one! It would certainly make review easier :)

techhat:
I'm not so worried about the number of commits, so long as it isn't outrageous. But other core devs will definitely come down hard on large amounts.

mafrosis:
@techhat I agree completely.

I think my reaction gets triggered whenever I see a commit that says "remove erroneously added file x".. Git rebase is a wonderful thing for borderline OCD people like me.

whodatninja:
Sorry about that. This should resolve the merge conflicts, and I cleaned up the commits.

techhat:
@whodatninja, sorry to take so long to look at this. Why are you adding the deploy_kwargs to the return? A couple of providers include a password in there, which some users may not want to display there... but the bigger problem is that one of the deploy_kwargs is the deploy script, which in the case of bootstrap-salt, will be _huge_. We definitely don't want that in there.

whodatninja:
@techhat, thanks for taking a look. Per some of the discussion last week in #403 I was seeking a way to pass back the assigned IP that would work consistently across existing cloud providers with minimal code impact. Also I was trying to avoid adding logic to cloud.create that referred too directly to run_map's implementation. In most cases, this approach required 2-ish lines to be edited/moved in cloud.create per provider. Thought was that to the extent it's memory consumptive, etc, (recognizing bootstrap-salt is 80kb) it's no more so than the cloud.create functions themselves, which will have just returned; and log impact could be addressed by removing deploy_args from output. 

A few strawman alternatives:
1. Remove deploy_args from run_map's return value as floated in #403 
2. Change cloud.creates to test for 'host' in deploy_args and, if present, add only that value to ret
3. Conclude that all approaches are too awkward relative to the use case here and scrap the idea
4. Others?

techhat:
My first inclination is to just sanitize deploy_args before adding it to the return; but our drivers (especially ec2) are getting more verbose in their return data, and this may just make it look like a mess. I'm still fence sitting on this one.

Let's not scrap the idea; people want it. I wonder if we can come up with something less awkward. I think it's time to look at integrating with overstate in Salt:

http://docs.saltstack.org/en/latest/ref/states/overstate.html

whodatninja:
Ok, fair enough. Happy to implement whatever you conclude is best.

With respect to overstate, I wasn't sure that it could address this use case. Isn't its whole design focused on the master executing sequenced state changes targeted at minions? In this use case, the master does not yet exist.

techhat:
@whodatninja, sorry to keep you waiting so long on this. It looks like, when using a map with this code, a "volumes" argument is required. This should not be the case; I don't know of anyone right now that is using volumes in their maps, so it definitely should not be required.

whodatninja:
Good catch. Should be fixed now.

techhat:
Thanks for your patience, @whodatninja. I'll test this tonight.

techhat:
Next problem:

  File "/usr/lib/python2.7/site-packages/saltcloud/cli.py", line 306, in run
    ret = mapper.run_map(dmap)
  File "/usr/lib/python2.7/site-packages/saltcloud/cloud.py", line 558, in run_map
    output[name] = self.create(tvm)
  File "/usr/lib/python2.7/site-packages/saltcloud/cloud.py", line 241, in create
    vm_['name']
ValueError: There's no master defined on the 'jhalltest' VM settings
Usage: salt-cloud

The jhalltest in the map does have a master defined. I don't know if this is related or not to your code... I'll do some poking around.

techhat:
Looks like the map_minion dict is not getting populated inside the tvm dict. This is in cloud.py, in Cloud.run_map(). Better make sure those map_ dicts are all getting set up right.

PR #442, opened 2013-03-13
example rename was not visible

ghost:
@techhat this is a good example of something I would just push directly, but PR's have more visibility and you might find that more valuable at this stage of salt-cloud. I generally haven't pushed much directly there so far except packaging tweaks.


techhat:
I could have sworn unlabeled code blocks worked, but this kind of thing keeps popping up this week, so I guess not. Thanks for letting me know.

PR #443, opened 2013-03-14
mention minion configs to disambiguate from grains example

ghost:
I don't mean to pick nits, but the example made me confused about what `minion:` is doing here.


inthecloud247:
good catch!

PR #445, opened 2013-03-14
Fix problems of not using location(ec2 region) defined in a VM profile during create()

kjkuan:
Tried salt-cloud with a fresh git clone, didn't work for me. These are the fixes that I had to make it work with my `cloud` and `cloud.profiles` configs. Basically, I'm using the `ec2` provider with a VM profile that specifies a location override:

<pre>
master:
  provider: ec2
  image: ami-be1230fb
  size: Micro Instance
  location: us-west-1
  ssh_username: ubuntu
  script: wget-bootstrap
  make_master: True
  minion:
    master: 127.0.0.1
    grains:
      cheese: edam
</pre>

I only did this for the ec2 provider, the `aws` provider might also have the same problem.
Please review and merge whatever that makes sense to you as I'm not yet familiar with the code base yet. Thanks!

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -165,8 +165,8 @@ def _xml_to_dict(xmltree):
     return xmldict
 
 
-def query(params=None, setname=None, requesturl=None, return_url=False,
-          return_root=False):
+def query(params=None, setname=None, requesturl=None, location=None,
+          return_url=False, return_root=False):
     key = __opts__['EC2.key']
     keyid = __opts__['EC2.id']
     if 'EC2.service_url' in __opts__:
@@ -175,8 +175,10 @@ def query(params=None, setname=None, requesturl=None, return_url=False,
         service_url = 'amazonaws.com'
     timestamp = datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
 
-    if not requesturl:
+    if not location:
         location = get_location()
+
+    if not requesturl:
         method = 'GET'
 
         if 'EC2.endpoint' in __opts__:
@@ -381,7 +383,7 @@ def script(vm_):
     minion = saltcloud.utils.minion_conf_string(__opts__, vm_)
     script = saltcloud.utils.os_script(
         saltcloud.utils.get_option(
-            'os',
+            'script',
             __opts__,
             vm_
         ),
@@ -410,24 +412,15 @@ def securitygroup(vm_):
 
 def ssh_username(vm_):
     '''
-    Return the ssh_username. Defaults to 'ec2-user'.
+    Return the ssh_username. Defaults to a built-in list of users for trying.
     '''
-    usernames = vm_.get(
-        'ssh_username', __opts__.get('EC2.ssh_username', 'ec2-user')
-    )
+    usernames = vm_.get('ssh_username', __opts__.get('EC2.ssh_username', []))
     if not isinstance(usernames, list):
-        username = usernames
-        usernames = [username]
-    if not 'ec2-user' in usernames:
-        usernames.append('ec2-user')
-    if not 'ubuntu' in usernames:
-        usernames.append('ubuntu')
-    if not 'admin' in usernames:
-        usernames.append('admin')
-    if not 'bitnami' in usernames:
-        usernames.append('bitnami')
-    if not 'root' in usernames:
-        usernames.append('root')
+        usernames = [usernames]
+    usernames = filter(lambda x: x, usernames) # get rid of None's or empty names
+    for name in 'ec2-user ubuntu admin bitnami root'.split():
+        if name not in usernames:
+            usernames.append(name)
     return usernames
 
 
@@ -578,7 +571,7 @@ def create(vm_=None, call=None):
         ).lower()
 
     try:
-        data = query(params, 'instancesSet')
+        data = query(params, 'instancesSet', location=location)
         if 'error' in data:
             return data['error']
     except Exception as exc:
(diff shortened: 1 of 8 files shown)


saltcloud/clouds/ec2.py
```diff
@@ -114,8 +114,9 @@ def __virtual__():
             )
         )
 
+#FIXME: WTF? why????
```
techhat: This is going away, hopefully today.

saltcloud/config.py
```diff
@@ -16,7 +16,7 @@
     'ssh_auth': '',
     'keysize': 4096,
     'os': '',
-    'script': 'bootstrap-salt',
+    'script': 'python-bootstrap',
```
techhat: This breaks compatibility with some non-Linux operating systems that don't have Python installed by default. This is one reason why we use bootstrap-salt in the first place.

techhat: Excellent example of this in:

https://github.com/saltstack/salt-cloud/blob/develop/saltcloud/deploy/SmartOS.sh

saltcloud/clouds/ec2.py
```diff
@@ -794,15 +788,17 @@ def set_tags(name, tags, call=None):
...
 
-    if 'Name' in tags:
-        return get_tags(tags['Name'], call='action')
+    #FIXME: we just set the tag, DescribeInstance may not have it yet!
+#    if 'Name' in tags:
+#        return get_tags(tags['Name'], call='action', location=location)
 
-    return get_tags(name)
+    # FIXME: if we ever reach here, this will result in sys.exit(1) since call != 'action'
+#    return get_tags(name)
```
techhat: I haven't seen it break yet, but it's easy enough to add 'call=action' here too...

saltcloud/clouds/ec2.py
kjkuan: Sorry, forgot to remove that. Please excuse the language, though it reflected my frustration at the time.

techhat: No worries. Frustrations abound in this specific area. I'm finishing pulling this out now.

saltcloud/clouds/ec2.py
kjkuan: I wasn't sure why `set_tags()` returns `get_tags(...)` since, at least in the `ec2` provider, the return value is never used. Another problem with calling `get_tags(...)` after setting the tag in `set_tags()` is that it seems the tags might not have been set. Finally, if we need to call `get_tags(...)` here then we should also pass it the `location`.

techhat: The idea here is to give feedback to the user, to let them know that their action has completed successfully. The set_tags() function shouldn't return until it gets a 200 OK status back from Amazon, and in theory, Amazon shouldn't be returning a 200 OK status until the operation has completed successfully. So by the time we get to get_tags(), it should be able to return the correct set of tags.

kjkuan: ok, so perhaps it was the missing `location` to `get_tags` that caused it to failed for me.

techhat: That wouldn't surprise me... there's some nuances with the location settings that are starting to bit us. Is it safe to assume you're not using us-east-1?

kjkuan: I used us-west-2.

saltcloud/clouds/ec2.py
techhat: Fixed in #446. This has likely caused a merge conflict here.

saltcloud/clouds/ec2.py
techhat: While I was doing #446, I also fixed these where I could find them. This will cause a merge conflict here as well, until your comments are removed.

techhat:
@kjkuan, I apologies for trampling on some of these with #446. I was already in the middle of it, and this one still needs some work. If you want to track me down as redbeard2 on irc, I can work with you on this.

kjkuan:
@techhat I just rebased my branch. Please review.

techhat:
Thank you, @kjkuan. It is mergeable now. I'll go ahead and check it out and run some tests.

saltcloud/utils/__init__.py
```diff
@@ -342,12 +345,12 @@ def deploy_script(host, port=22, timeout=900, username='root',
             # Run the deploy script
             if script:
                 log.debug('Executing /tmp/deploy.sh')
-                if make_syndic:
-                    deploy_command += ' -S'
-                if make_master:
-                    deploy_command += ' -M'
                 if 'bootstrap-salt' in script:
-                    deploy_command += ' -c /tmp/'
+                    deploy_command += ' -c /tmp/'  #FIXME: always?
```
techhat, line 349: Since this only happens for bootstrap-salt, and those files are used for make_master, yes. Good call on moving the other arguments in here.

saltcloud/clouds/ec2.py
```diff
@@ -960,14 +950,33 @@ def show_instance(name, call=None):
...
+def _vm_provider(vm_):
+    return vm_.get('provider', __opts__['provider'])
+
+
+def _list_nodes_full(location=None):
+    ret = {}
     params = {'Action': 'DescribeInstances'}
-    instances = query(params)
+    instances = query(params, location=location)
+#    import pprint
```
techhat: Did you mean to leave the pprint in here?

kjkuan: no, I left it there by accident.

techhat:
@kjkuan, I have finished my testing, and it's looking good. Would you like to remove those pprint lines before I merge? Also, would you like to add yourself to AUTHORS?

kjkuan:
@techhat, thanks for reviewing. I've removed the `pprint`. Adding to AUTHORS is not necessary, thanks!

techhat:
Thanks @kjkuan, you're teh awesome!

PR #446, opened 2013-03-14
Update create_attach_volumes() in EC2 to use direct API

techhat:
This removes the final vestiges of libcloud support from this driver. If a user would like to use libcloud, the AWS driver still uses it.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -115,9 +115,6 @@ def __virtual__():
             )
         )
 
-    # open a connection in a specific region
-    conn = get_conn(**{'location': get_location()})
-
     log.debug('Loading EC2 cloud compute module')
     return 'ec2'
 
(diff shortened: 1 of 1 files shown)


saltcloud/clouds/ec2.py
```diff
@@ -706,32 +681,60 @@ def create(vm_=None, call=None):
 
     volumes = vm_.get('map_volumes')
     if volumes:
-        log.info('Create and attach volumes to node {0}'.format(data.name))
-        create_attach_volumes(volumes, location, data)
-
+        log.info('Create and attach volumes to node {0}'.format(vm_['name']))
+        import pprint
+        pprint.pprint(ret['placement']['availabilityZone'])
```
s0undt3ch: you really want pprint here?

```
log.info(ppring.pformat(ret['placement']['availabilityZone']) 
```

techhat: How embarrassing; I thought I got all those. Fixed.

PR #448, opened 2013-03-15
Add EC2 to "new config" dicts

techhat:
This may resolve #439. @fredvd, would you please verify?


techhat:
Apologies, @fredvd, this will not resolve #439.

PR #451, opened 2013-03-18
Openstack ignore configured cidr for private addresses

boltronics:
My company has an internal OpenStack install for development purposes and what not. Each instance we launch with salt-cloud has two IP addresses - one on our main internal network accessible by employees (where our salt master currently is located) which uses a floating IP address range, and the other is a flat network IP which is not visible to the salt master.

Unfortunately salt-cloud usually tries to execute the bootstrap script by connecting on the wrong IP, and there appears to be no way to prevent this... until now. The pull request here allows something like OPENSTACK.ignore_cidr: 192.168.40.0/24 to be set so we can now have our backend flat network ignored and the correct IP is always chosen.

As simple as it is, probably the Salt devs could do a nicer job of this code than I, so please do clean it up if/where appropriate. I'm not a huge fan of bringing in another Python dependency (python-netaddr), but I'm not sure if there's a better way. Thanks!!

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,6 +19,7 @@ Depends: ${python:Depends},
          ${shlibs:Depends},
          python,
          python-libcloud,
+         python-netaddr,
          python-yaml,
          salt-common,
          sshpass
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,6 @@ Depends: ${python:Depends},
          ${shlibs:Depends},
          python,
          python-libcloud,
-         python-netaddr,
          python-yaml,
          salt-common,
          sshpass
--- a/doc/man/salt-cloud.7
+++ b/doc/man/salt-cloud.7
@@ -575,6 +575,16 @@ OPENSTACK.apikey: 901d3f579h23c8v73q9
 .ft P
 .fi
 .sp
+If your OpenStack instances only have private IP addresses and a CIDR range of
+private addresses are not reachable from the salt-master, you may set your
+preference to have Salt ignore it:
+.sp
+.nf
+.ft C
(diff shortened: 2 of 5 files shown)


techhat:
Thanks for the pull req @boltronics. I'm not a fan of bringing in another dependency either, so I will investigate as to whether we can avoid it. At the very least, since this dep is only for Openstack, it should not be included in the requirements.txt or the debian control file; instead, documentation should be added to Openstack concerning installing it.

Just another quick note: there's no need to modify man pages; they get autogenerated (overwritten) when a release is made, so any changes made to them directly will be discarded.

I'll take a look at this today and see if we can work around that dep. In the meantime, you will want to consider backing out those other things I mentioned.

boltronics:
Thanks @techhat for the quick response and those pointers, and for looking into this. I've backed out those changes to the man and control files, and noted the netaddr dependency within the OpenStack module.

techhat:
@boltronics, let me ask one more thing before I merge this. Take a look at the yumpkg module in Salt:

https://github.com/saltstack/salt/blob/develop/salt/modules/yumpkg.py

There is a try/except import block towards the top that sets a HAS_YUMDEPS flag. You could something like use this in your code to make the ignore_ip_addr() function only work if netaddr has been installed.

With that change in place, I would accept this code.

techhat:
@boltronics, I have merged this manually with the change that I suggested. Thanks for the work!

boltronics:
Thanks @techhat. Sorry for not getting around to doing it myself. I've been flat out on another project and had to put it on hold. Much appreciated.

techhat:
Not a problem @boltronics, thanks again!

PR #454, opened 2013-03-20
Check if the 'append_domain' option is set before accepting the key.

ConsoleCatzirl:
If the minion is configured with the 'append_domain' option, then we need to add that string to the key name before accepting the generated key for the minion.  Otherwise, when the minion starts it will generate a new key request which needs to be manually accepted, and a stale accepted request will go unused.


techhat:
FYI, man pages are auto-generated from the .rst files when a new release is cut, so any changes made will be overwritten.

PR #457, opened 2013-03-23
Renamed the default includes to be less confuse.

s0undt3ch:
- `/etc/salt/cloud.d` was renamed to `/etc/salt/cloud.conf.d`
- `/etc/salt/profiles.d` was renamed to `/etc/salt/cloud.profiles.d`


techhat:
Better.

techhat:
Addresses #347.

PR #462, opened 2013-03-25
Try a few times before failing when getting node data using the EC2 driver

s0undt3ch:
Try a few times before failing when getting node data using the EC2 driver


inthecloud247:
Why not change the delay to 1 second per attempt, and extend to 60 total attempts. The current defaults seems low, no? It currently only attempts for 5 seconds total.

PR #463, opened 2013-03-25
Only delete a machine once(when both aws and ec2 is enabled)

s0undt3ch:
- Only delete a machine once(when both aws and ec2 is enabled).
- No need to create unused vars.


techhat:
I think we need to take a different approach here. This only takes into account aws and ec2, and that won't work when we start allowing multiple configurations for the same provider.

s0undt3ch:
And it was only handling `destroy`

PR #464, opened 2013-03-25
Fix list locations for Linode

jasonjohnson:
Resolves https://github.com/saltstack/salt-cloud/issues/452

This may be an issue for other providers. Was only able to test Linode.


techhat:
You're the man, @jasonjohnson.

PR #469, opened 2013-03-26
A few bug fixes:

kjkuan:
- Fix incorrect ordering of arguments to the -f option.
  BTW, personally, I think using this format:
  
  ```
  -f &lt;provider&gt;.&lt;function&gt; [name1=value name2=value2 ...]
  ```
  
  makes more sense than the current format:
  
  ```
  -f &lt;function&gt; &lt;provider&gt; [name1=value name2=value2 ...]
  ```
  
  but anyways.
- Give better error message when -p <profile_name> is used but no instance
  names are supplied.
- Make ec2's list_node() also available for use with the -f option.
- Fix the incorrect assumption that a minion dict will always be
  available in the config when removing minion keys from master.
  It's possible that the node is a salt-master node created by the
  make_master option.

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -242,8 +242,8 @@ class SaltCloud(parsers.SaltCloudParser):
 
         elif self.options.function:
             prov_func = '{0}.{1}'.format(
-                self.function_name,
                 self.function_provider,
+                self.function_name
             )
             if prov_func not in mapper.clouds:
                 self.error(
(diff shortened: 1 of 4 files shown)


techhat:
I just changed it a couple of weeks ago to this format:

```
salt-cloud -f <function name> <provider>
```

to match the functionality of actions:

```
salt-cloud -a <action name> <instance>
```

What is the reasoning for changing it back?

kjkuan:
The current format is fine. This patch doesn't change it, please ignore my rant :)  This patch only fixes it since it's not parsing the command line correctly.

s0undt3ch:
Some of the fixes brought in this PR are already handled or are going to be handled by other PR's.
Regarding the order change, this order was adopted by a reason, that reason is still valid.

Sorry @kjkuan.

PR #472, opened 2013-03-26
Switch to `urllib2`. Sometimes `urllib` returns `None` when calling `getcode()`

s0undt3ch:
Switch to `urllib2`. Sometimes `urllib` returns `None` when calling `getcode()`


techhat:
Yes, urllib2 is much more reliable... if only it did _all_ of the things that urllib does.

PR #483, opened 2013-03-30
EC2 instance provisioning is incomplete and does not set Name tag (https://github.com/saltstack/salt-cloud/issues/478)

toppur:
A query to list EC2 instances might return either a dict or a list (for stopped instances). Need to handle both cases. See the issue cited above for example error messages in the log .


techhat:
This is much cleaner, @toppur, thanks.

PR #496, opened 2013-04-04
Add Saltify driver

techhat:
Fix #493.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -841,9 +841,14 @@ def del_tags(name, kwargs, call=None):
         log.error('The del_tags action must be called with -a or --action.')
         sys.exit(1)
 
+    if not 'tags' in kwargs:
+        log.error('A tag or tags must be specified using tags=list,of,tags')
+        sys.exit(1)
+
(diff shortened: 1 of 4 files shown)


saltcloud/clouds/saltify.py
```diff
@@ -0,0 +1,176 @@
...
+# Import salt libs
+import salt.utils.event
+import salt.utils.xmlutil
+import saltcloud.utils
+
+# Get logging started
+log = logging.getLogger(__name__)
+
+
+# Only load in this module if the PARALLELS configurations are in place
```
s0undt3ch, line 33: This comment could be removed?

saltcloud/clouds/ec2.py
```diff
@@ -841,9 +841,14 @@ def del_tags(name, kwargs, call=None):
         log.error('The del_tags action must be called with -a or --action.')
         sys.exit(1)
 
+    if not 'tags' in kwargs:
```
s0undt3ch, line 844:
``` python
if 'tags' not in kwargs:
```

s0undt3ch:
@UtahDave that finger is too close to the trigger :smile:

PR #498, opened 2013-04-04
Add support for block device mappings in AWS.

codysoyland:
This adds support for AWS block device mappings (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html)

This requires libcloud >= 0.12.1 (http://svn.apache.org/viewvc/libcloud/trunk/libcloud/compute/drivers/ec2.py?r1=1438350&r2=1438349&pathrev=1438350)

### DIFF
--- a/doc/topics/aws.rst
+++ b/doc/topics/aws.rst
@@ -194,6 +194,18 @@ Multiple security groups can also be specified in the same fashion:
       - default
       - extra
 
+Block device mappings enable you to specify additional EBS volumes or instance
+store volumes when the instance is launched. This setting is also available on
+each cloud profile:
+
+.. code-block:: yaml
+
+    AWS.block_device_mappings:
+        - DeviceName: /dev/sdb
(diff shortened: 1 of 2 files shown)


techhat:
Thanks for the pull req, @codysoyland! I would like to direct your attention to #450, and #121.

codysoyland:
Thanks, techhat! It looks like the functionality in #121 is to create and attach EBS volumes to running instances. If I understand correctly, you can only attach ephemeral instance store volumes to new instances on launch, and that is why you must supply a block device mapping upon instance launch. In addition to allowing you to attach ephemeral drives, this patch also allows you to specify volumes as part of your cloud profile script.

techhat:
Ah! My apologies, I didn't read it closely enough, it would seem. This is most excellent, and it makes me want to update the changes in #121 to make it more clear, like this is. I'll get this merged now, and we can fix that later.

PR #501, opened 2013-04-07
Added the LON variant identity_url

binocarlos:
I was plugging in salt-cloud with a UK Rackspace account - after a while fumbling around as to why it wasn't liking me - I realised the authentication URL has lon prepended.

I figured a small note in the docs would be useful for the next guy : )


techhat:
Hey, thanks for the note, @binocarlos! Getting it in there.

@s0undt3ch, I don't know if this will affect your changes; you may need to rebase. But it's tiny.

PR #521, opened 2013-04-10
Allow the download of the bootstrap script to be skipped. Fixes #517.

s0undt3ch:
Either passing `--skip-bootstrap-download` to `setup.py build` or by having `SKIP_BOOTSTRAP_DOWNLOAD=1` as an environment variable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ import os
 import urllib2
 from distutils import log
 from distutils.core import setup
-from distutils.command.build import build as distutils_build
+from distutils.command.sdist import sdist as original_sdist
 
 setup_kwargs = {}
 USE_SETUPTOOLS = False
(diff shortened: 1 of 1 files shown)


s0undt3ch:
@techhat, @herlo, @seanchannel, @aboe76: The bootstrap script download automation is now **only** done when running `python setup.py sdist`.

What this means is:
- if you're creating a distro package with a source tarbal, everything is included;
- if you're creating a distro package with a git checkout, then the bootstrap script needs to be downloaded and put in the proper place by hand, as if `salt-cloud -u` was executed.
- An alternative is to change the download code to a custom distutils command which get's called by `sdist`. This last approach allows you to do `python setup.py <custom bootstrap download command>` after a git checkout, or, `python setup.py <custom bootstrap download command> build compile install`.

What's the preferred and less problematic way of doing this, taking into consideration the particularities of each build system?

techhat:
I'm still in favor of putting the bootstrap script directly in the repo. This will help daily PPAs get built, and @s0undt3ch's work will still save me a step when building a release tarball.

aboe76:
@s0undt3ch, @herlo, @seanchannel, @techhat : could this be salt-cloud 0.8.8?

Or do a backport the setup.py and ship a modfied salt-cloud-0.8.7.tar.gz to keep package version in sync
with all distro's ?

s0undt3ch:
Oh well, add the script to the repo. Guess most package related problems will be gone.

@aboe76: Regarding a backport or a 0.8.8 release, it's up to @techhat

PR #541, opened 2013-04-19
#540 hardcoded to use public ip for deploy.sh ssh calls

rszabo50:
I have introduced the code to check the vm configuration for the key ssh_interface. 

If set to priviate_ips, it will use the private ip to deploy, which is intended for when the salt-master/salt-cloud server is inside the data-center.

eg:

joyent-centos-small:
    provider: joyent-sw-cloud
    image: centos-6
    ssh_interface: private_ips
    script: RHEL6
    size: Extra Small 512 MB


techhat:
Excellent! Thanks, @rszabo50!

PR #543, opened 2013-04-20
#542 - adding support for reboot action

rszabo50:
added new method to joyent.py to support reboot operations currently available in libcloud


techhat:
Huh, I had thought that we had already added reboot to JoyEnt. I must have been thinking of a different provider. Thanks for getting these in here, @rszabo50!

techhat:
I wanted to point out that I like what you're doing with has_method(). I wonder if that should be moved into libcloudfuncs.py? It's good stuff.

rszabo50:
I will put it in there now.

PR #557, opened 2013-04-25
Fixed bug where minion name was indexing a string instead of a dict

pandu-rao:
I found this bug when I tried to create a map to instantiate a set of nodes. The minion name was being used as an index when the variable miniondict was a string. I indented it to the correct level and salt-cloud was able to proceed to create the nodes based on the map definition.

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -838,9 +838,9 @@ class Map(Cloud):
                             tvm['map_minion'] = miniondict[name]['minion']
                         if 'volumes' in miniondict[name]:
                             tvm['map_volumes'] = miniondict[name]['volumes']
-                for myvar in miniondict[name]:
-                    if myvar not in ('grains', 'minion', 'volumes'):
-                        tvm[myvar] = miniondict[name][myvar]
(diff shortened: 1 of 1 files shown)


gilclark:
This will fix issue #547.   So +1.

gilclark:
Even though I don't think I have a vote :-).

jcollie:
After applying this patch, I get the following traceback:

[root@home01 ~]# salt-cloud -m /etc/salt/cloud.map
[INFO    ] Configuration file path: /etc/salt/cloud
[INFO    ] Applying map from '/etc/salt/cloud.map'.
The following virtual machines are set to be created:
  svr03.ocjtech.us

Proceed? [N/y] y
... proceeding
[ERROR   ] There was a query error: 'map_minion'
  File "/usr/lib/python2.7/site-packages/saltcloud/cli.py", line 277, in run
    ret = mapper.run_map(dmap)
  File "/usr/lib/python2.7/site-packages/saltcloud/cloud.py", line 845, in run_map
    tvm['minion'] = tvm['map_minion']
KeyError: 'map_minion'

PR #568, opened 2013-05-06
fixing private IPs for openstack

nevins-b:
This changes fixes and issue where data.private_ips was not being set if public_ips where provided. It resolves issue [199](https://github.com/saltstack/salt-cloud/issues/199).

### DIFF
--- a/saltcloud/clouds/openstack.py
+++ b/saltcloud/clouds/openstack.py
@@ -410,11 +410,15 @@ def create(vm_):
                 not_ready = False
                 break
 
+        if running and private:
+            data.private_ips = private
+            if ssh_interface(vm_) == 'private_ips':
+                not_ready = False
+
         if running and public:
             data.public_ips = public
-            if private:
-                data.private_ips = private    
-            not_ready = False
+            if ssh_interface(vm_) != 'private_ips':
+                not_ready = False
 
         if not_ready is False:
             break
--- a/saltcloud/clouds/openstack.py
+++ b/saltcloud/clouds/openstack.py
@@ -412,6 +412,8 @@ def create(vm_):
 
         if running and public:
             data.public_ips = public
+            if private:
+                data.private_ips = private    
             not_ready = False
 
         if not_ready is False:


saltcloud/clouds/openstack.py
```diff
@@ -412,6 +412,8 @@ def create(vm_):
 
         if running and public:
             data.public_ips = public
+            if private:
```
s0undt3ch: Shouldn't this code be:

``` python
if running and public:
    data.public_ips = public
if running and private:
    data.private_ips = private
```

nevins-b: I don't think so, I think if anything you may want it to look at ssh_interface(vm_) and see if it should continue waiting for private IPs once it has public ones. I would think 90% of the time people will want to code to return after it has the public IPs.

s0undt3ch: My change does not add to the wait, I think.

s0undt3ch: Ah, sorry, I removed the `not_ready = False`

s0undt3ch:
``` python
if running and public:
    data.public_ips = public
    not_ready = False
if running and private:
    data.private_ips = private
```

?

s0undt3ch: Like the above, even if public ips is not know already, and private is, it's set. Break out of the loop when the public ip is found?

s0undt3ch:
``` python
if running and public:
    data.public_ips = public
    not_ready = False
if running and private:
    data.private_ips = private
    if ssh_interface(vm_) == 'private_ips':
        not_ready = False
```

This should be it.

s0undt3ch: Want to update the code so you can still own the change wrights? Just push the new change to your repository and it should show up here.

nevins-b:
Hmm I think what we actually will want is:

``` python
if running and private:
    data.private_ips = private
        if ssh_interface(vm_) == 'private_ips':
            not_ready = False

 if running and public:
     data.public_ips = public
         if ssh_interface(vm_) != 'private_ips':
             not_ready = False
```

This will set both private and public IPs as well as waiting for private to be set if ssh_interface(vm_) == 'private_ips'

s0undt3ch:
Thanks @nevins-b!

PR #571, opened 2013-05-09
Updated the Compute region about LONDON data center

Katafalkas:
In documentation it was not mentioned that you also have to specify the OPENSTACK.compute_region: LON for UK based RackSpace accounts to work


s0undt3ch:
Thanks @Katafalkas!

PR #578, opened 2013-05-11
#577 - API Update Changes and removal of libcloud dependency

rszabo50:
#577

Fixes issues with latest release of the API

### DIFF
--- a/saltcloud/clouds/joyent.py
+++ b/saltcloud/clouds/joyent.py
@@ -15,8 +15,6 @@ Using the old cloud configuration syntax, it requires that the ``username`` and
     JOYENT.password: saltybacon
     # The location of the ssh private key that can log into the new VM
     JOYENT.private_key: /root/joyent.pem
-    # the Datacenter location associated with the new VMS
-    JOYENT.location: us-east-1
 
 Using the new format, set up the cloud configuration at
  ``/etc/salt/cloud.providers`` or ``/etc/salt/cloud.providers.d/joyent.conf``:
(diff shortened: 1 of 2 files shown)


saltcloud/libcloudfuncs.py
```diff
@@ -332,3 +337,19 @@ def conn_has_method(conn, method_name):
         )
     )
     return False
+
+def is_private_subnet(ip):
```
techhat: Two comments here. First, only functions specific to libcloud belong in this file. Second, this duplicates functionality which already existts in the is_public_ip() function in utils/**init**.py.

saltcloud/clouds/joyent.py
```diff
@@ -139,44 +157,37 @@ def create(vm_):
     Create a single VM from a data dict
     '''
 
+    #pprint(inspect.stack())
```
techhat: Forgotten?

saltcloud/clouds/joyent.py
```diff
@@ -255,74 +281,81 @@ def create(vm_):
     return ret
 
 
-def reboot(name, call=None):
-    '''
-    Reboot a node.
+def create_node(**kwargs):
```
techhat, line 291: This is one of many functions that doesn't contain docstrings. Each function needs at least a one line description, though a CLI Example is not necessarily required unless usage isn't obvious. With all the new additions lately, that is likely to change (meaning CLI Examples will probably be required soon).

rszabo50:
i will make those changes tomorrow and you can review again

saltcloud/libcloudfuncs.py
rszabo50: agreed, and did not find that function earlier. will change and test tomorrow

saltcloud/clouds/joyent.py
rszabo50: needs to be removed, used it to find full stack trace when debugging an earlier issue

rszabo50:
this is ready for review

PR #585, opened 2013-05-17
Ec2 subnetid

binarymatt:
Adds the SubnetId and SecurityGroupId params to the ec2 cloud driver for vpc support. Fixes issue issue #440

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -783,7 +783,7 @@ def create(vm_=None, call=None):
             'Returned query data: {0}'.format(data)
         )
 
-        if 'ipAddress' in data[0]['instancesSet']['item']:
+        if 'ipAddress' in data[0]['instancesSet']['item'] or 'privateIpAddress' in data[0]['instancesSet']['item']:
             # We have our IP, break out of the loop
             break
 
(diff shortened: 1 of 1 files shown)


saltcloud/clouds/ec2.py
```diff
@@ -242,7 +242,7 @@ def query(params=None, setname=None, requesturl=None, location=None,
         params['SignatureVersion'] = '2'
         params['SignatureMethod'] = 'HmacSHA256'
         params['Timestamp'] = '{0}'.format(timestamp)
-        params['Version'] = '2010-08-31'
```
techhat, line 245: Do we know for sure that we're 100% compatible with the 2013 version of the spec?

binarymatt, line 245: i don't. I upgraded that in order to use the SecurityGroupId parameter. I've been running this on my salt-master and haven't seen any issues, but it could be backed out pretty easily. Would just need to remove said SecurityGroupId params

saltcloud/clouds/ec2.py
```diff
@@ -592,6 +592,19 @@ def get_availability_zone(vm_):
 
     return avz
 
+def get_subnetid(vm_):
```
techhat, line 595: If you wouldn't mind adding docstrings (using single quotes, not double quotes) to these couple of functions, I think I'm okay with merging this.

binarymatt:
added the doctstrings to the get_subnetid and securitygroupid functions

techhat:
Thanks duder! We'll need to get some docs in the aws.rst file as well, but this will be good for now.

binarymatt:
i'll try to get some docs written up this weekend.

PR #588, opened 2013-05-19
Make use of a common IP waiting function for all cloud providers

s0undt3ch:
Make use of a common IP waiting function for all cloud providers


s0undt3ch:
This pull request is not yet complete. I'll update the information here when it's ready for a review.

s0undt3ch:
@techhat this is ready for review.

PR #594, opened 2013-05-23
Adding CloudStack provider

sebgoa:
This creates a basic cloudstack provider. This was tested.
What's missing is the key based authentication to the instances.

### DIFF
--- a/saltcloud/clouds/cloudstack.py
+++ b/saltcloud/clouds/cloudstack.py
@@ -1,0 +1,216 @@
+'''
+CloudStack Cloud Module
+===================
+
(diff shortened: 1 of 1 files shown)


techhat:
Awesome, @runseb! Question: what do host and path refer to?

sebgoa:
host and path define your cloudstack endpoint, it's the variables necessary for the cloudstack driver in libcloud.
For instance for an endpoint: http://localhost/api you would have.
host='localhost'
path='/api'

techhat:
Excellent. We should put together a quick-start guide for cloudstack, but this is already good stuff!

PR #596, opened 2013-05-24
Build the providers map in parallel using multiprocessing.

s0undt3ch:
This way we can narrow down the map building time since some of them are try/fail query attempts against the remote API.


techhat:
So this will always do map lookups in parallel, but not necessarily do anything else in parallel, right?

s0undt3ch:
yes, always maps in parallels, nothing else in parallel, unless `-P` is passed

PR #597, opened 2013-05-24
Handles IPs better to be compatible with libcloud

sebgoa:
libcloud patch has been accepted but we need this new version in saltstack to be more compatible.

### DIFF
--- a/saltcloud/clouds/cloudstack.py
+++ b/saltcloud/clouds/cloudstack.py
@@ -130,6 +130,18 @@ def get_keypair(vm_):
 
     raise SaltCloudNotFound('The specified keypair could not be found.')
 
+def get_ip(data):
+    '''
+    Return the IP address of the VM
+    If the VM has  public IP as defined by libcloud module then use it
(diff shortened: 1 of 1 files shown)


techhat:
I've been watching the traffic for this on libcloud, and can appreciate where you're coming from on that side of things. Is this the best way to handle this on the salt-cloud side of things? Compare and contrast with:
- https://github.com/saltstack/salt-cloud/blob/develop/saltcloud/clouds/ec2.py#L814-L819
- https://github.com/saltstack/salt-cloud/blob/develop/saltcloud/clouds/openstack.py#L386-L416

sebgoa:
Right now, I am going for quick and working, rather than clean and complete :)
I test things against a working public cloud, which means that the code may be a bit tailored to that cloud. For instance there is no concept of publicly addressable IPs on that cloud, the private IPs in libcloud lingos are actually public ones that Salt-cloud needs.

Once the dust settles, I will re-visit things 

-sebastien

techhat:
Sounds like a plan. Thanks @runseb!

PR #602, opened 2013-06-01
If there is no Name tag, just return instance id

WillPlatnick:
If a vm has no name tag but has multiple tags, salt-cloud failed to return a name, which messed up being able to query correctly.


techhat:
Wait, why did you close this?

WillPlatnick:
I did it wrong :)

PR #603, opened 2013-06-01
If an item has a list of tags but no Name, return the instanceId

WillPlatnick:


WillPlatnick:
Fixes https://github.com/saltstack/salt-cloud/issues/600

PR #623, opened 2013-06-09
Fix bug in wait_for_ip: ValueError: Unknown format code 'd' for object of type 'float'

ageron:
(timeout // 60)  is an integer even when timeout is a float.  But (timeout % 60) is a float when timeout is a float, so it cannot be formatted with "...{1:02d}...".format(...).
So I changed it to int(timeout % 60).

Same for duration % 60.


ageron:
(timeout // 60) actually returns a float too, so changed it to int(timeout // 60).

s0undt3ch:
Ooops...

Thanks @ageron

PR #630, opened 2013-06-12
Added option to skip ssl verification for openstack

KennethWilke:
With this addition a user can add 'insecure: True' to the openstack provider config to skip certificate validation

fixes #629

### DIFF
--- a/saltcloud/clouds/openstack.py
+++ b/saltcloud/clouds/openstack.py
@@ -193,6 +193,14 @@ def get_conn():
         )
     }
 
+    insecure = config.get_config_value(
+        'insecure', vm_, __opts__, search_global=False
+    )
+
+    if insecure:
+        import libcloud.security
+        libcloud.security.VERIFY_SSL_CERT = False
+
     password = config.get_config_value(
         'password', vm_, __opts__, search_global=False
     )


s0undt3ch:
Hmm, `insecure` might not be the best config var name, perhaps `verify_ssl_cert`? Seems explicit enough.

KennethWilke:
I went with insecure as that's what's used in novaclient, the primary application i use to interact with openstack outside of salt, but skip_ssl_cert sounds good as well.

s0undt3ch:
Well, I suggested that because it needs less documentation since it's quite explicit :smiley:

thatch45:
I would lean towards keeping config values consistent with the underlying provider, but on the other hand there is the question of making options easier to translate to other providers as well. with that said, I don't have a side to take...

s0undt3ch:
Also, please reset the setting after done with it. This won't matter with CLI usage but makes a difference with API level usage.

KennethWilke:
Where would be a good location to place code to reset the setting?

s0undt3ch:
Something similar would work:

``` python
    if insecure:
        import libcloud.security
        previous_verify_ssl_cert = libcloud.security.VERIFY_SSL_CERT
        libcloud.security.VERIFY_SSL_CERT = False

    # cut code ------------
    ret = driver(
        config.get_config_value('user', vm_, __opts__, search_global=False),
        config.get_config_value('apikey', vm_, __opts__, search_global=False),
        **authinfo
    )


    if insecure:
        libcloud.security.VERIFY_SSL_CERT = previous_verify_ssl_cert

    return ret
```

techhat:
FWIW, my vote is to continue call this insecure, since that's what it's already called in Salt's own OpenStack modules:

https://github.com/saltstack/salt/blob/develop/salt/modules/keystone.py

s0undt3ch:
Your vote rules :)

@KennethWilke just reset the setting and we're good to merge it

KennethWilke:
s0undt3ch: No dice on that recommendation :( libcloud only allows certificate verification on a global level, so it'd need to be reset somewhere else.

s0undt3ch:
@KennethWilke do you get a traceback with the ssl failure?

s0undt3ch:
@KennethWilke no worries. I'm merging this and I'll open a ticket so we don't forget the API usage and this global setting.

s0undt3ch:
We'll take care of the setting reset issue in #632

s0undt3ch:
@KennethWilke Thanks!

KennethWilke:
@s0undt3ch thank you!

PR #655, opened 2013-06-23
Updated the code to check again for the public IP address of an instance...

amosdavis:
... instead of only checking once.  The one time the code checks is not sufficient to get a public IP address from a newly created instance.  This code change has been tested against HP Cloud Services openstack implementation.


s0undt3ch:
Thanks @amosdavis!

PR #670, opened 2013-06-28
Fixes issue #669

abemusic:
Allows snapshot to be provided in volume definitions so the snapshot may be used in create_attach_volume. If snapshot is not provided it defaults back to using the size attribute.


techhat:
@abemusic, you are the awesome!

abemusic:
Thanks @techhat  :+1:

PR #694, opened 2013-07-29
Dependencies in cloud map

mattmb:
Hi All,

Here's what I've been working on as part of the salt sprint.

I wanted to be able to make machines dependent on each other in a cloud map so that they get deployed in a specific order.

So you can do something like:

```
ubuntu_small:
    - web01:
        requires:
            - db01
    - web02:
        requires:
            - db01
    - db01
    - loadbalancer:
        requires:
            - web01
            - web02
```

This is straightforward for non parallel deployments because I can just order the list before we deploy. However it is more tricky in parallel. I found that adding a start_action for parallel deployments wasn't working at all so with these changes the start actions get triggered after all the machines have been deployed (but respecting the dependencies). So the machines are all created at the same time but the start actions run in the correct way afterwards.

I'd welcome some feedback/debate on whether this is the best way of doing it? Also on formatting the start_action output?

Thanks,
Matt

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -946,6 +946,42 @@ class Map(Cloud):
                 map_[profile][name] = {'name': name}
         return map_
 
+    def _has_loop(self, dmap, seen=None, val=None):
+        if seen is None:
+            for k,v in dmap['create'].items():
+                seen = []
+                try:
+                    machines = v['requires']
+                except KeyError as e:
+                    machines = []
+                for machine in machines:
+                    if self._has_loop(dmap, seen=list(seen), val=machine):
+                        return True
+        else:
+            if val in seen:
+                return True
+            else:
+                seen.append(val)
+                try:
+                    machines = dmap['create'][val]['requires']
+                except KeyError as e:
+                    machines = []
+                for machine in machines:
+                    if self._has_loop(dmap, seen=list(seen), val=machine):
+                        return True
+        return False
+
+    def _calcdep(self, dmap, machine, data, level):
+        level+=1
+        try:
+            for name in data['requires']:
+                data = dmap['create'][name]
+                level = self._calcdep(dmap, name, data, level)
+            return level
+        except KeyError:
+            level-=1
+            return level
+
     def map_data(self, cached=False):
         '''
         Create a data map of what to execute on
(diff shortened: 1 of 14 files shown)


techhat:
@s0undt3ch, I'm looking at this, but could I get your eyes on it as well?

mattmb:
Just noticed that I hadn't thought about maps where some of the machines already exist (D'oh). Also I'm not raising an error if you depend on a machine that isn't in the map. Expect another commit to fix!

techhat:
Make sure to make a note when you do so, so I don't miss it.

saltcloud/cloud.py
```diff
@@ -613,7 +614,15 @@ def create(self, vm_, local_master=True):
...
-
+        # If it's a map then we need to respect the 'requires'
+        # so we do it later
+        if not self.opts['map']:
+            if self.opts['parallel'] and self.opts['start_action']:
+                client = salt.client.LocalClient()
+                out = client.cmd(
+                    vm_['name'], self.opts['start_action'], timeout=300
+                )
+                pprint.pprint(out)
```
s0undt3ch: Can you please switch the `pprint.pprint` calls with salt's outputter system? That was the user is able to choose which output format he wishes.

s0undt3ch: For an example for the outputter system, look [here](https://github.com/mattmb/salt-cloud/blob/454b33fc7d1e7edf5e785be83e7fc4d4203eede5/saltcloud/cli.py#L111).

Also, no output should be done in `saltcloud.cloud`. `saltcloud.cloud` should return data which should be displayed in `saltcloud.cli`. This allows proper API access to salt-cloud.

s0undt3ch:
Salt recently got ordered states. We can implement the same for maps.
So, the challenge at hand is parallel, so, unordered deployment for which start actions should be ordered?

Sorry if I narrowed it down too much, just woke up :)

mattmb:
So I've fixed my first effort so that it will now handle when some hosts already exist. It will now also raise an error if you require something not in the map. Finally I've put the output from the start_action into the output dictionary. Although I haven't changed anything for non-parallel deployments.

mattmb:
@s0undt3ch do you mean ordered states as in explicitly including an order i.e.

```
- host:
    - order: 2
```

or are you saying we should order maps in the same way that salt orders states? Presently my code only works with 'requires'.

mattmb:
Any update on this? Is there anything else you'd like me to do to get this functionality in?

techhat:
Sorry for the lag, @mattmb. This seems to be working well in my testing. Thanks for doing this!

PR #702, opened 2013-08-04
Removing references to old configuration format from EC2

techhat:
It's time to start pulling references for the old configuration format from the documentation. This is the first step.


s0undt3ch:
:+1:

PR #714, opened 2013-08-10
Add support for IOPS provisioned volumes

mahendra:
Also ensure that the volumes detail is updated in the documentation


techhat:
Thanks, @mahendra! FYI, man pages are automatically generated for releases, so there's no need to parch them here.

mahendra:
@techhat thanks for merging :-)

PR #716, opened 2013-08-10
Update to regions, adding SYD, removing old auth endpoint info for UK/LON

welbymcroberts:
Rackspace now have a region in Sydney, and have merged the identity_url endpoints, I've updated the docs to reflect this and highlighted that UK accounts are the only ones to have access to the LON region, and have no access to other regions

### DIFF
--- a/doc/topics/rackspace.rst
+++ b/doc/topics/rackspace.rst
@@ -66,31 +66,16 @@ that Salt Cloud has been built to support.
 Compute Region
 ==============
 
-Rackspace currently has three compute regions which may be used:
+Rackspace currently has four compute regions which may be used:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


techhat:
Is the LON region not required to be specified for UK accounts then, as it is the default? I seem to recall people adding those notes because of problems they were having in not specifying them, but I could be misremembering.

welbymcroberts:
A region is always required to be set, for both UK and other accounts. Up until about a month ago the UK was using a different set of auth endpoints, and due to this needed the identity_url to be set, the region was still required then. Now that the auth endpoints have merged, the identity_url is no longer required.

techhat:
Ah, I think I understand. Thanks @welbymcroberts!

PR #717, opened 2013-08-10
Fix bug in cli argument parsing

mahendra:
The argument list was being manipulated inside the for loop.
This was causing the loop to behave very strangely on some systems.
The loop would exit after the first iteration and the command would
exit with an error

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -252,7 +252,7 @@ class SaltCloud(parsers.SaltCloudParser):
         elif self.options.function:
             kwargs = {}
             args = self.args[:]
-            for arg in args:
+            for arg in args[:]:
                 if '=' in arg:
                     key, value = arg.split('=')
                     kwargs[key] = value


saltcloud/cli.py
```diff
@@ -252,7 +252,7 @@ def run(self):
         elif self.options.function:
             kwargs = {}
             args = self.args[:]
```
s0undt3ch, line 254: `args` is already a copy of `self.args`.

Can we get the tracebacks/errors you were getting please?

mahendra, line 254: yep, i agree. we are using args as a copy of self.args so that self.args is preserved.

but the args list is being modified during it's iteration, (which I think is causing the problem). The log is the one caused by the next line

``` python
if args:
    self.error('Any arguments passed to --function need to be passed ' ...)
```

mahendra, line 254: @s0undt3ch let me know if you need any more details.

s0undt3ch, line 254: A failing example would be awesome :)

techhat:
Hi @mahendra, have you been able to get an example yet?

mahendra:
@techhat sorry for the delay. Here it is:

```
LC02KV0PBFFT3:outfiles mahendram$ sudo salt-cloud  --profiles=/tmp/outfiles/cloud.profiles.d/us-east-1.conf  --providers-config=/tmp/outfiles/cloud.providers.d/us-east-1.conf -f create_volume ec2 zone=us-east-1b size=10 type=standard
Password:
[INFO    ] Configuration file path: /etc/salt/master
[INFO    ] There is no AWS cloud provider configuration available. Not loading module
Usage: salt-cloud

salt-cloud: error: Any arguments passed to --function need to be passed as kwargs. Ex: image=ami-54cf5c3d. Remaining arguments: ['size=10']
LC02KV0PBFFT3:outfiles mahendram$
```

mahendra:
@techhat BTW, I am running this on OSX. Python 2.7.2

I agree that we are doing args = self.args[:], but we are modifying args inside the iteration. I think that is what is causing this strange behavior. Basically, in my case, the loop exited after the first iteration itself. (verified this with print statements)

UtahDave:
@techhat does @mahendra's example help here?

s0undt3ch:
The idea that we're changing the args in the iteration is what's supposed to happen. This way we know if extra, non handled args are being passed.

@mahendra a traceback would really be very helpful here, I'm still not convinced, and believe me, I'll be the first to apologize if you show me the light that I'm failing to see...

techhat:
Just talked to @s0undt3ch about this. @mahendra's change will keep us from modifying the list as we're looping through it (which is good, because you should never modify the thing that you are looping through, as you're looping through it).

s0undt3ch:
@mahendra so, the first issue I saw today was this one, and I don't know if that made any difference but I saw something today which I had not seen before.
So, as I said, :bow:, **I'm sorry for the time it took for me to see the light, you are correct**!  We we're iterating through a list which we were changing, as you also pointed out. I wonder why this didn't blew up before...

s0undt3ch:
Oh, @techhat also explained it :smile:

PR #718, opened 2013-08-12
Ensure that all volumes created during init follow delvol_on_destroy

mahendra:
If a user uses salt-cloud --map and make an instance. He uses the "volumes" option and create/attach volumes to this instance (with delvol_on_destroy = True in the configuration)

Now, if the instance is destroyed using salt-cloud -d --map option. In such a situation, the non-root volumes must also be destroyed.

Basically when the instance is created the delvol_on_destroy option must be set on all volumes created by the salt-cloud script.

Also, the delvol/keepvol_on_destroy actions must work on any volume attached to the instance.

The question also comes up - what will happen to a volume if it is detached from the instance before the instance is destroyed. In such a case, it is the admin's decision. The volume will not be destroyed.


techhat:
Let me take a closer look at this when I get back to my computer.

techhat:
@mahendra, I understand the importance of what you're trying to do, but I don't think we should change the current expected behavior to accomplish it. I think I would rather deprecate the existing option name, and add an option for `del_root_vol_on_destroy`, which performs identically (probably using an alias). Then we could also add an option for `del_attached_vols_on_destroy`, which performs as you need.

What are your thoughts on that?

mahendra:
@techhat I agree to that point. makes more sense.

mahendra:
@techhat have split out the options as you asked for.

techhat:
@mahendra, I was just about to merge, and I saw a merge conflict. I think we're good to go on this as soon as that's cleared up. Would you mind rebasing your branch?

mahendra:
@techhat thanks. have rebased it

techhat:
Thanks @mahendra, you're awesome!

mahendra:
@techhat thanks for reviewing :-)

actually, I am an idiot. I had not tested the patch thoroughly. If there were more than one volumes attached to the device, the last one was not respecting the flag. Also kwargs was not being followed properly for the commands/actions for setting or removing this flag. Have sent you another pull request for this.

Sorry for the trouble.

PR #719, opened 2013-08-14
Ensure hostnames are set properly. 

mahendra:
Should fix #712

### DIFF
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -413,6 +413,9 @@ def deploy_script(host, port=22, timeout=900, username='root',
                 root_cmd(subsys_command, tty, sudo, **kwargs)
                 root_cmd('service sshd restart', tty, sudo, **kwargs)
 
+            # Update hostname on the minion
+            root_cmd('test `hostname` == {0} || hostname {0}'.format(name))
+
             # Minion configuration
             if minion_pem:
                 scp_file('/tmp/minion.pem', minion_pem, kwargs)
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -414,7 +414,8 @@ def deploy_script(host, port=22, timeout=900, username='root',
                 root_cmd('service sshd restart', tty, sudo, **kwargs)
 
             # Update hostname on the minion
-            root_cmd('test `hostname` == {0} || hostname {0}'.format(name))
+            hostname_cmd = 'test `hostname` == {0} || hostname {0}'.format(name)
+            root_cmd(hostname_cmd, tty, sudo, **kwargs)
 
             # Minion configuration
             if minion_pem:


techhat:
Hi @mahendra. This change is a start, and I think it's important so I'll merge it... but it's not a persistent fix. If the node reboots, the fix goes away. However, since you're pretty unlikely to be rebooting right away (and may never reboot an instance before destroying it), let's get it in there. Which distro are you running?

mahendra:
@techhat thanks for the merge, but I am confused as to how the changes will be lost. The hostname command will set it to /etc/hostname, right? This will persist among reboots most of the time.

Also, the real hostname of the minion is not important to salt. It is only needed to be correct during the key generation and enrollment with salt-master. After that salt will not worry about it, right?

techhat:
@mahendra just as a sanity check, I tried this on an ec2 box (Ubuntu 13.04) and it did not persist to /etc/hostname.

Since hostnames can be so unreliable, salt does not technically care about them. You can set the minion ID manually in /etc/salt/minion, and then salt will go from there. If you don't set it manually, salt will default to the hostname... and if some random process decides to change the hostname and then salt (re)starts, it will effectively invalidate the minion's key on the master, because salt will identify as the new hostname. For this reason, it's always best to just manually specify the minion ID manually (which salt-cloud should be doing for you anyway).

mahendra:
@techhat in that case, we can update the minion information for each minion with the vm['name'] in <tt>deploy_script</tt>, right? will that solve the problem? if so, i can submit a patch for it

techhat:
@mahendra you shouldn't need to go to all that trouble. Just set the minion options that you want in your configuration and/or map file:

http://salt-cloud.readthedocs.org/en/latest/topics/config.html#minion-configuration
http://salt-cloud.readthedocs.org/en/latest/topics/map.html

Is there something else that you're needing to do that I'm missing?

mahendra:
yep, that will work. However I like to avoid setting information in duplicate. the minion id is now being set twice. Was looking at avoiding that (so that it does not lead to any mistakes - specially with copy-paste :-)

Also, the patch has a problem. Setting hostname has to be 'sudo test .... || sudo hostname <a....>'

techhat:
Nope, you did it right. If you look more closely at the root_cmd() function, you'll see that it handles sudo logic for you. On some providers sudo is required, and on (most) others, a root login is used, and sudo is not required. You can override the sudo option in your configs, but for most instances, the correct default is already figured out for you.

mahendra:
@techhat I meant the second sudo. In some cases, sudo "command" >> /abc would not work. Only the command is run under sudo. The > or >> operators don't get the permission. I am working on another ticket to make the hostname changes permanent. hostname command only invokes the sethostname system call, which is why it disappears on reboot. You need to set it manually in /etc/hostname and /etc/hosts. Will send you a diff in a short while after testing it out.

techhat:
@mahendra: Aha! I understand. this is an important use case, thanks for bringing it up. Maybe it would make sense to allow for a flag to be used, which would be interpolated the same as sudo in the root_cmd() function.

A workaround for now might be:

```
(test `hostname` == {0} || hostname {0})
```

(using parens to enclose the command)

s0undt3ch:
Regarding output redirection and piping and perms, google got `tee`.

s0undt3ch:
errr, google **for** `tee`.

mahendra:
sure thanks. plz do have a look at #722 - just ensured that the changes are permanent.

WillPlatnick:
FYI, this commit broke all of my salt-cloud VM's. When I sudo, it throws back text that it can't resolve the hostname because the hostname isn't in my /etc/hosts. This makes capistrano not able to sudo correctly.

techhat:
@tempspace this has now been reverted, are your instances behaving better now?

WillPlatnick:
I ended up commenting out the changes and built my own package that day and
everything was fine.

PR #722, opened 2013-08-15
Ensure that hostname setting is permanent

mahendra:
Fixes issues raised in #712 and #719

### DIFF
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -422,12 +422,12 @@ def deploy_script(host, port=22, timeout=900, username='root',
                 root_cmd(hostname_cmd, tty, sudo, **kwargs)
 
                 # Update the hostnames file
-                hostname_cmd = 'sh -c "echo -n {0} > /etc/hostname"'.
+                hostname_cmd = 'sh -c "echo -n {0} > /etc/hostname"'.\
                                     format(name)
                 root_cmd(hostname_cmd, tty, sudo, **kwargs)
 
                 # Update /etc/hosts
-                hosts = 'sh -c "echo -e "127.0.0.1\t{0}\t{1}" >> /etc/hosts"'.
+                hosts = 'sh -c "echo -e "127.0.0.1\t{0}\t{1}" >> /etc/hosts"'.\
                                 format(name.split('.')[0], name)
                 root_cmd(hosts, tty, sudo, **kwargs)
 
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -422,8 +422,7 @@ def deploy_script(host, port=22, timeout=900, username='root',
                 root_cmd(hostname_cmd, tty, sudo, **kwargs)
 
                 # Update the hostnames file
-                hostname_cmd = 'sh -c "echo -n {0} > /etc/hostname"'.\
-                                    format(name)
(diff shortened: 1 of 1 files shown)


saltcloud/utils/__init__.py
```diff
@@ -414,8 +414,21 @@ def deploy_script(host, port=22, timeout=900, username='root',
...
+            if root_cmd(test_hostname, tty, sudo, **kwargs) != 0:
+                # Set the hostname
+                hostname_cmd = 'hostname {0}'.format(name)
+                root_cmd(hostname_cmd, tty, sudo, **kwargs)
+
+                # Update the hostnames file
+                hostname_cmd = 'sh -c "echo {0} > /etc/hostname"'.format(name)
+                root_cmd(hostname_cmd, tty, sudo, **kwargs)
+
+                # Update /etc/hosts
```
techhat, line 428: This looks like it has the possibility of creating duplicate lines in `/etc/hosts`. I'm not sure this is the correct way to go about this.

saltcloud/utils/__init__.py
```diff
@@ -414,8 +414,21 @@ def deploy_script(host, port=22, timeout=900, username='root',
...
-            root_cmd(hostname_cmd, tty, sudo, **kwargs)
+            test_hostname = 'test `hostname` == {0}'.format(name)
+
+            if root_cmd(test_hostname, tty, sudo, **kwargs) != 0:
+                # Set the hostname
+                hostname_cmd = 'hostname {0}'.format(name)
+                root_cmd(hostname_cmd, tty, sudo, **kwargs)
+
+                # Update the hostnames file
+                hostname_cmd = 'sh -c "echo {0} > /etc/hostname"'.format(name)
```
techhat, line 425: Are you sure the sh -c is necessary here? I would think that echo on its own should be enough... plus, in the non-Linux distros, I don't know that sh can reliably be expected to exist (which sounds silly, but it's true).

mahendra, line 425: oh! will try without -c and see how it goes. Hmmm, I thought salt was only for Linux distros :-) (kidding)

Maybe, we can move this whole logic into a shell script, scp it to the box and run it with sudo. That might help. Any other ideas?

saltcloud/utils/__init__.py
mahendra, line 428: You mean two lines with 127.0.0.1 in it?

<pre>
127.0.0.1 localhost
127.0.0.1 mybox.mydomain.com
</pre>


Is that a problem? It works fine, AFAIK

techhat:
@mahendra, I'm trying to figure out your purpose here. You're going to a lot of work to set the hostname bit, when I don't think it's really necessary for you. Based on other conversations with you, it sounds like your major concern is setting the minion ID properly, right? This should be set automatically based on the name you gave the minion when you spun it up in the first place. Is that not happening for you?

mahendra:
@techhat that is the problem. it was not happening. That is why I opened ticket #712

s0undt3ch:
I'm getting to this thread a bit late here. I really don't think setting the hostname is salt-cloud's job.

You set a proper minion ID to the machine to be created, boostrap salt, and once salt is running, salt provides you the tools to mess with your hostname, which, does not necessarily need to be equal to the hostname.

s0undt3ch:
As stated in #747, the minion ID's are properly being set and `salt-cloud` should have nothing to do with the VM's hostname. Closing. Please re-open if need be.

PR #740, opened 2013-08-20
Adds spot instance support to EC2 driver

abemusic:
For each VM in your map, add

``` json
"spot_config": {
    "max_price": "0.1"
}
```

 where `max_price` is the maximum price you're willing to pay for the spot instance. For more information on EC2 spot instances, see http://aws.amazon.com/ec2/spot-instances/


techhat:
@abemusic, you are the man!

abemusic:
@techhat haha, thanks...glad I can help :)

lusid:
There should be a spot_prefix on the two IamInstanceProfile mappings as well. Without them, a spot instance cannot be created with an IAM role assigned.

PR #747, opened 2013-08-28
Fix for #744 and some improvements

s0undt3ch:
- If the master failed to create, do not continue with the map.
- When search for configuration keys on the VM's configuration or it's provider or the global config, don't forget that VM's profile configuration. When bootstrapping VMs using the profile this extra step is not necessary, though, when bootstrapping using maps, the map only holds the profile name, not it's data, making this extra step, fundamental.

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -1251,6 +1251,12 @@ class Map(Cloud):
                     )
                 )
 
+            elif 'Errors' in out:
+                raise SaltCloudSystemExit(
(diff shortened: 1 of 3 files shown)


techhat:
Tagging @mahendra.

s0undt3ch:
Since during the tests the minion ID's were properly set, I also reverted the changes from https://github.com/saltstack/salt-cloud/pull/719, `salt-cloud` should have nothing to do with the bootstrapped machine's hostname.

PR #750, opened 2013-08-29
Suppress SSH output

pabelanger:
A simple patch to reduce the noise on the CLI. For example:

Warning: Permanently added 'a.b.c.d' (ECDSA) to the list of known
hosts.

### DIFF
--- a/saltcloud/utils/__init__.py
+++ b/saltcloud/utils/__init__.py
@@ -727,7 +727,9 @@ def scp_file(dest_path, contents, kwargs):
         # Set hosts key database path to /dev/null, ie, non-existing
         '-oUserKnownHostsFile=/dev/null',
         # Don't re-use the SSH connection. Less failures.
-        '-oControlPath=none'
+        '-oControlPath=none',
+        # Don't be verbose when running commands.
+        '-o LogLevel=quiet',
     ]
     if 'key_filename' in kwargs:
         # There should never be both a password and an ssh key passed in, so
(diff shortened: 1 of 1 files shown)


techhat:
@pabelanger, while I know that some users (including yourself) will benefit from this, I believe it will cause problems for others (including myself).

I think the better solution here is to allow users to pass custom options directly to the SSH command, and set the default for this particular option to `quiet` as you have done. Then developers such as myself can run in verbose mode as needed.

pabelanger:
I agree, I was thinking more about it last night before bed.  Going to close it out for now until I can get a better option.

krak3n:
Was anything done Re the SSH output? I'ts quite annoying the json output from the -p command also contains all the SSH output, it means I can't pragmatically load the stuff I want into a json parser as it's all munged together, `-l quiet` doesn't seem to do the job.

techhat:
@krak3n nothing was from this pull req, but there is an option to suppress SSH output, that was added in #385 (display_ssh_output).

If you're interested in the develop branch, we have merged salt-cloud into salt, which paved the way for a lot of new usages of salt-cloud. For instance, it can now be used as a salt execution or state module, or it can be called from the new CloudClient API.

krak3n:
Thanks @techhat, yeah im very much looking forward to Hydrogen's inclusion of Salt Cloud :+1:. 

`display_ssh_output` is exactly what I need :)

PR #756, opened 2013-08-30
Added verify_ssl_cert and networkid options

pakdel:
verify_ssl_cert can be used to disabled SSL certificate verification, which
is useful in case a self-signed certificate is used on CloudStack server.
networkid can be used to launch the server into a specific network.
Also it uses extra['password'] that is provided by latest version of libcloud.

As a result, I get to have cloud.providers and cloud.profiles files like the following:
cloud.providers 

```
mycloud:
  apikey: "Some Key"
  secretkey: "Some Key"
  host: "cloud.ops.dot"
  path: "/client/api"
  provider: cloudstack
  verify_ssl_cert: false
```

cloud.profiles

```
small:
    provider: mycloud
    image: CentOS 6 minimal
    size: Small Instance
    location: YUL01
    ssh_interface: public
    keypair: mycloud
    networkid: ab815684-880b-4c45-91ce-8e8041d15e63
```

### DIFF
--- a/saltcloud/clouds/cloudstack.py
+++ b/saltcloud/clouds/cloudstack.py
@@ -27,6 +27,8 @@ import logging
 import saltcloud.config as config
 from saltcloud.libcloudfuncs import *   # pylint: disable-msg=W0614,W0401
 from saltcloud.utils import namespaced_function
+# CloudStackNetwork will be needed during creation of a new node
+from libcloud.compute.drivers.cloudstack import CloudStackNetwork
 
 # Get logging started
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


techhat:
Thanks for the contribution, @pakdel!

pakdel:
Wow! Great! That was the fastest merge!

pakdel:
Should I do anything to make sure that these changes will be present in next releases?

techhat:
Nope, they're in the develop branch, so they'll be in the next release.

pakdel:
I cloned the develop branch today, and it seems that this commit has been only merged into 0.8.9!

techhat:
This commit shouldn't even be in 0.8.9, since it wasn't added until after that release was cut. It should only be in develop at the moment.

techhat:
It has just barely been pointed out to me that I merged this into the 0.8.9 branch, which is incorrect. It should have been merged into develop only. @s0undt3ch will submit the correct pull req(s) to fix this.

pakdel:
I created a pull request for develop branch: #773

PR #765, opened 2013-09-02
Add show_delvol_on_destroy and other bug fixes

mahendra:
Add a <tt>show_delvol_on_destroy</tt> command/action to see the status of the flag.

There were some bugs in #718
- If you attach 1 or more volumes (using <tt>volumes</tt>), the last volume to be attached will not have the flag set. This was because of a loop issue where the <tt>_toggle_delvol</tt> was being invoked before the <tt>attach</tt> API was being invoked.
- The <tt>delvol_on_destroy</tt> and <tt>keepvol_on_destroy</tt> commands were not respecting the keyword arguments (was resulting in 400/Bad request)

### DIFF
--- a/doc/topics/aws.rst
+++ b/doc/topics/aws.rst
@@ -485,6 +485,7 @@ using one of the following commands:
 
     salt-cloud -a delvol_on_destroy myinstance
     salt-cloud -a keepvol_on_destroy myinstance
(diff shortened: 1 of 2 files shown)


mahendra:
phew! thanks :-)

PR #772, opened 2013-09-05
Added verify_ssl_cert and networkid options

s0undt3ch:
verify_ssl_cert can be used to disabled SSL certificate verification, which
is useful in case a self-signed certificate is used on CloudStack server.
networkid can be used to launch the server into a specific network.


techhat:
Thanks, @s0undt3ch!

techhat:
Affects #756.

PR #774, opened 2013-09-06
Various small fixes

mafrosis:
Pull request include several smallish fixes I've been using over the few days:
- 7e1d51d 60 seconds for a socket to timeout is rather long - especially since we then retry the socket for a total of 15 minutes.
- 594ac9a is what should've been done in e101032.
- 0400e0c exit the CLI main method in a single place on success.
- 8282bca display errors in the right place.
- 1d41cec return existing instances in output on map create - this has the added benefit of preventing salt-cloud chucking an error code when running a map file where all nodes already exist (which I believe is not an error state).

### DIFF
--- a/saltcloud/cli.py
+++ b/saltcloud/cli.py
@@ -114,11 +114,7 @@ class SaltCloud(parsers.SaltCloudParser):
         if self.selected_query_option is not None:
             if self.selected_query_option == 'list_providers':
                 try:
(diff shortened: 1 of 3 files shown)


s0undt3ch:
I agree with all of these, Thanks @mafrosis!!!

mafrosis:
@s0undt3ch glad to hear it!

PR #777, opened 2013-09-06
Fixing small issue in DescribeInstances result handler

abemusic:
Attempting to reference the `error` key in an empty list.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -1020,10 +1020,8 @@ def create(vm_=None, call=None):
 
         if isinstance(data, list) and not data:
             log.warn(
-                'There was an error in the query. {0} attempts '
-                'remaining: {1}'.format(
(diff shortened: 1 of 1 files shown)


techhat:
Is the list always empty? If there is sometimes an error in there, I would like to see it. Can we instead do something like:

```
err_msg = ''
if 'error' in data:
    err_msg = data['error']
```

abemusic:
As the code was written, it was testing for an instance of a list and that the list was empty. In that case `data['error']` would always throw:

``` bash
TypeError: list indices must be integers, not str
```

I'm not sure why the query result would be an empty list though.

techhat:
Sorry, I should have looked closer before opening my big mouth. You are correct.

abemusic:
@techhat No worreis...I had to take a second look myself to make sure I wasn't doing something silly :-D

PR #781, opened 2013-09-10
digital ocean: added support for new private networking flag

sibsibsib:
I've added support for Digital Ocean's new private networking flag during node creation. This code checks against the region id to make sure it matches NY2, since this feature is only available there. Available region is encapsulated in a function so that new ones can be added as DO expands. 

The flag can be set in a profile like such:

```
ubuntu_ocean_tiny:
    provider: dig-ocean
    image: Ubuntu 12.04 x64
    size: 512MB
    private_networking: True
```

When the node is deployed, you should see a new field in the api response called `private_ip_address`.

### DIFF
--- a/saltcloud/clouds/digital_ocean.py
+++ b/saltcloud/clouds/digital_ocean.py
@@ -124,6 +124,13 @@ def avail_sizes():
     return ret
 
 
+def avail_private_networks():
+    '''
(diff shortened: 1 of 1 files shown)


techhat:
Thanks @sibsibsib, you're teh awesome!

PR #787, opened 2013-09-17
[openstack] additionnal for argument for libcloud to work with openstack...

arthurzenika:
... essex (closes #698)

### DIFF
--- a/doc/topics/config.rst
+++ b/doc/topics/config.rst
@@ -466,6 +466,17 @@ Using the new syntax:
       ignore_cidr: 192.168.0.0/16
 
 
+For in-house Openstack Essex installation, libcloud needs the service_type :
+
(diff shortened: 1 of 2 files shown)


techhat:
This is close, @arthurlogilab, but there is a change that needs to be made here. In this code, service_type will always be passed in, and we want it to be optional, kind of like what we did with the password.

arthurzenika:
With https://github.com/arthurlogilab/salt-cloud/commit/f5ea4648623f11db3d94113a7d65edd948a3994c which makes it optional, is this pull request OK ?

techhat:
Apologies, @arthurlogilab, I didn't see that commit come in; unfortunately, github doesn't notify me when commits are added to a pull req.

PR #788, opened 2013-09-17
EC2: add constraint when querying for IP address

malept:
When querying for an IP address after instance creation, only return the
private IP address if it's asked for in the profile. This tends to be a
problem when dealing with instances in VPCs, and it takes a while for
the public IP address to be generated.

Fixes saltstack/salt-cloud#693.


techhat:
I think I can get behind that.

PR #802, opened 2013-09-19
SoftLayer Hardware Cloud fixes

underscorephil:
Added in a fix for ordering and an object mask to reduce the number of calls necessary for information gathering.


techhat:
You rock, @underscorephil!

PR #803, opened 2013-09-19
[SoftLayer] Avoids deprecated uses of SoftLayer Client

sudorandom:
Here is a reference for deprecated APIs in the SoftLayer python client: http://softlayer-api-python-client.readthedocs.org/en/latest/api/client/#backwards-compatibility

This change also fixes some formatting issues and removes unused imports for the SoftLayer drivers.


techhat:
Oh man, thanks for cleaning this up, @sudorandom!

PR #809, opened 2013-09-23
Allowing optional_products to be specified for softlayer-hw

techhat:


techhat:
Tagging @underscorephil.

PR #813, opened 2013-09-24
Added IAM Profile for EC2 driver, and a warning on AWS Profile

johnnoone:
IAM roles is a nice feature for restricting instances permissions.
http://docs.aws.amazon.com/IAM/latest/APIReference/Welcome.html

I added a new key on EC2 provider:

```
EC2.iam_profile: "iam profile name or iam profile arn"
```

Now, about AWS driver.

While AWS is based on `libcloud`, and it does implement this, currently I raise an Exception, inviting people to try EC2 driver.

I've also pulled a request to `libcloud` https://github.com/apache/libcloud/pull/150 . When it will be accepted, we will remove the exception and finish the implementation.

### DIFF
--- a/saltcloud/clouds/ec2.py
+++ b/saltcloud/clouds/ec2.py
@@ -562,6 +562,30 @@ def securitygroup(vm_):
     )
 
 
+def iam_profile(vm_):
+    '''
+    Return the IAM profile.
+
+    The IAM instance profile to associate with the instances.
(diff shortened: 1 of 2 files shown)


techhat:
Thanks, @johnnoone, it looks good. I did see the libcloud pull req as well, and am looking forward to seeing it merged.

What would the setting look like in the configs? Would you be willing to include an example in the docs in your next pull req?

PR #814, opened 2013-09-24
Added some explanation for IAM profile & roles

johnnoone:
voilà !
it's not the most clear explanation but I it's enough for AWS gurus.


techhat:
Excellent, thanks @johnnoone!

PR #815, opened 2013-09-24
Add the ability to specify VLAN when spinning up an instance

techhat:


techhat:
Tagging @underscorephil.

PR #816, opened 2013-09-24
Add the ability to set max network speed on softlayer virtual guests

techhat:


techhat:
Tagging @underscorephil.

PR #823, opened 2013-09-24
Update example with correct provider

duncanmurray:
Updating example with new format to include correct provider.

### DIFF
--- a/doc/topics/rackspace.rst
+++ b/doc/topics/rackspace.rst
@@ -59,7 +59,7 @@ that Salt Cloud has been built to support.
       tenant: 123456
       apikey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
-      provider: rackspace
+      provider: openstack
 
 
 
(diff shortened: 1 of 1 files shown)


techhat:
@duncanmurray, you missed a couple. Did you want me to wait on this so you can make them consistent?

duncanmurray:
Updated the ones I missed. Should be all the references covered. Also added the IAD region as well

techhat:
@duncanmurray, not seeing other commits.

doc/topics/rackspace.rst
```diff
@@ -59,14 +59,14 @@ that Salt Cloud has been built to support.
       tenant: 123456
       apikey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
-      provider: openstack
+      provider: rackspace
```
techhat: This is wrong. The rackspace driver will choke on this configuration. The correct driver name for this is openstack.

PR #824, opened 2013-09-25
Change wait_for_ssh to wait_for_port

techhat:
This makes more sense, in terms of salting non-Unix/Linux systems, such as Windows.


techhat:
Tagging @UtahDave.

PR #825, opened 2013-09-25
ec2 set_tags was failing on integer values and values with spaces

jstrunk:
This fixes a bug I found. I didn't create an issue since the fix was so obvious.

Thank you!


techhat:
You're awesome, @jstrunk!

jstrunk:
Thank you for the kind words.

I just found another case that this fix doesn't help with: dates. I need to see if there are any options to make config.get_config_value() interpret the YAML as a dict, but get the values as strings.

So, expect another pull request soon.

PR #827, opened 2013-09-25
check for strings in tag dict's values. Give a helpful error message.

jstrunk:
The issue is that tags in map files are being interpreted as python data types when they really need to be uninterpreted strings. For example, a date gets changed to a datetime object, but that is not useful.

The workaround is to quote the values in the map file.

I think the issue of yaml values getting interpreted as python data types is deeper in salt's code than I should change without understanding the wider ramifications. So, I have added some more input checking to warn people if their tag values are getting changed. It suggests the workaround of adding quotes.


techhat:
I'm down with that.

PR #828, opened 2013-09-26
use more robust heredoc syntax for updating the minion configuration

ewdurbin:
it appears that at some point configuration values containing ' snuck into the minion config. this patch will handle unescaped characters like that and many more


techhat:
Word.

PR #835, opened 2013-09-29
Add support for OpenStack floating IPs

pabelanger:
This attempts to fix #637 and fix #791. However, this code currently
depends on libcloud 0.14.0-dev, which has not been released yet.  I have
updated the requirements for tox to handle the development release of
libcloud however I don't expect it to get merged.

So, for now, this is just a public patch for people to test out.  I'll
revist the commit message once the next release of libcloud is out.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 # pip requirements file for Salt Cloud
 salt
-apache-libcloud
+-e git+https://github.com/apache/libcloud.git#egg=apache-libcloud-0.14.0-dev
--- a/saltcloud/clouds/openstack.py
+++ b/saltcloud/clouds/openstack.py
@@ -63,7 +63,10 @@ configuration at ``/etc/salt/cloud.providers`` or
       ssh_key_name
       # The OpenStack network UUIDs
       networks:
-          - 4402cd51-37ee-435e-a966-8245956dc0e6
+          - fixed:
+              - 4402cd51-37ee-435e-a966-8245956dc0e6
+          - floating:
+              - Ext-Net
 
       provider: openstack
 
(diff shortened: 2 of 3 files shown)


techhat:
The changes to requirements.txt are not going to fly. Libcloud is required all over salt-cloud, and cannot be removed from requirements.txt. netaddr is only used in openstack, and should not be in requirements.txt.

Still looking at the rest of the changes.

pabelanger:
libcloud is still a requirement, I have just bumped it to 0.14.0-dev.  For this patch to work, 0.14.0 will have to be the minimum requirement.  I am trying to find out when the next release will be.

As for netaddr, I believe it needs to be in requirements, since salt-cloud uses it.  I can move that into a separate patch.

The other changes are to ignore_cidr, that code was a mess and I cannot see how it ever worked.

techhat:
Some of the cloud modules have requirements that only apply to them; softlayer and softlayer-hw require the SoftLayer package. The legacy aws driver required botocore. And the openstack driver is the only thing that requires netaddr. Just like in Salt proper, we don't add packages to requirements.txt that are only required by one module; among other reasons, it places undue stress on our community packagers (we learned that the hard way with botocore). That being the case, netaddr doesn't go into requirements.txt.

pabelanger:
Okay, I guess there is a difference on how you guys use the requirements file, in the past any dependency that an application needed to running correctly would be placed their [1]. However, I can move it into test-requirements.txt because it will be needed if unit tests will be needed.

Additionally, if you want to keep separate requirements for setup.py and tox, that can be done too.  However, will just lead to duplicate pip install files.

[1] http://www.pip-installer.org/en/latest/cookbook.html#requirements-files

PR #837, opened 2013-09-29
Add support for userdata to OpenStack

pabelanger:
A simple patch to allow a user to pass userdata when a node is created
for OpenStack.


pabelanger:
The more I think about this, I think it should support both inline syntax or reading an external file.

pabelanger:
Decided to change it to an external file, it should give the user more options.

techhat:
@pabelanger what is the file format like for the included file? Is it based on an existing format from openstack?

pabelanger:
@techhat I'd say the format should be per http://cloudinit.readthedocs.org/en/latest/ however, it really depends on what is reading the data.  The choice is really up to the user.

techhat:
@pabelanger looks good, thanks!

PR #844, opened 2013-10-03
[debian] add python-netaddr as a Recommended package for OpenStack cloud providers

arthurzenika:


techhat:
Thanks @arthurlogilab!

PR #846, opened 2013-10-04
Ported block_device_mappings to ec2 provider

clintonb:
This functionality (which allows for ephemeral storage setup) was previously only available in the AWS provider. Documentation has also been updated to prevent confusion when setting up block devices on unsupported instance types.

 Fixes #845

### DIFF
--- a/doc/topics/aws.rst
+++ b/doc/topics/aws.rst
@@ -324,7 +324,10 @@ Multiple security groups can also be specified in the same fashion:
 
 Block device mappings enable you to specify additional EBS volumes or instance
 store volumes when the instance is launched. This setting is also available on
-each cloud profile.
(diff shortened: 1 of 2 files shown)


clintonb:
Let me know if there are tests I need to update. I didn't see any in /tests.

techhat:
@clintonb, just a note before I review this, it looks like your commit is showing up as a ghost commit. This can happen when the email addressed that it was committed under isn't registered with Github. I can still merge it, but until that address is registered, Github won't give you credit.

clintonb:
@techhat Thanks. My profile is now up to date.

techhat:
@clintonb looks good, both the profile and the code. Thanks for doing this!

PR #850, opened 2013-10-07
Add support for Rackspace Managed service level

duncanmurray:
This patch will wait for the Rackspace managed kick script to complete before continuing with the bootstrapping. Doing this will stop the setup of the minion from also attempting to install packages while the Rackspace install process is still running. Completion of the process is determined by checking the value of the "rax_service_level_automation" key.


techhat:
@duncanmurray, I would like to merge this, but there's a merge conflict. Could you rebase, please?

duncanmurray:
@techhat openstack.py should now be current

techhat:
Looks great @duncanmurray, thanks!

PR #856, opened 2013-10-15
Ensure that instance creation works without root volumes

mahendra:
Some instances do not have a root volume. In such cases, if we have
del_root_volume_on_destroy enabled, salt-cloud will fail. Ideally
it should ignore this setting


techhat:
You're awesome as always, @mahendra!

mahendra:
@techhat thanks for the merge :-)

PR #859, opened 2013-10-16
Fix cloudstack driver to handle self-signed certs, unsecure connection a...

sebgoa:
...nd better handling of keypairs and network_id


techhat:
Thanks @runseb!

PR #860, opened 2013-10-17
Add Windows support to all current cloud providers

techhat:
This even adds it to Linode, which doesn't technically support Windows, and may not ever do so. But if they do, we'll be ready!


techhat:
Bam!

s0undt3ch:
:bomb:

underscorephil:
Nice!

PR #868, opened 2013-10-21
Initial commit of nova driver

techhat:
DO NOT MERGE

This contains some initial work on an openstack driver using the novaclient and glanceclient python bindings. It currently exists here for discussion.


techhat:
I'm going to merge what I have now, so that others can start hacking on this. I have added the proper warnings concerning its stability to the docstring.

PR #870, opened 2013-10-22
15 seconds was too short a timeout. Leave default

UtahDave:


UtahDave:
@s0undt3ch could you take a peek at this?  This timeout defaults to 15 seconds, which was way too short or DigitalOcean.  Leaving the default timeout allows DigitalOcean to work.

PR #876, opened 2013-10-24
Moved disabling SSL verification from __virtual__() to get_conn()

whiteinge:
**virtual**() is called once when the code is first loaded and won't
necessarily have the right context if there is more than one
configuration entry for the same provider.

All glory goes to @s0undt3ch for figuring out the problem.

### DIFF
--- a/saltcloud/clouds/cloudstack.py
+++ b/saltcloud/clouds/cloudstack.py
@@ -63,20 +63,6 @@ def __virtual__():
         )
         return False
 
-    verify_ssl_cert = config.get_config_value(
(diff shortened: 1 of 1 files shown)


whiteinge:
Thanks for the fix! You are very handsome and have incredibly smart friends. ;-)

whiteinge:
Thank you for the quick merge. Yes, my friends are very smart indeed.

s0undt3ch:
LOLOLOLOLOL!!!!!!!

You lost it? :nut_and_bolt:

PR #877, opened 2013-10-25
Syntax and formatting fixes

s0undt3ch:


whiteinge:
Egad, I copy-and-pasted two-space indents without noticing. I'm going to head home to shower...

PR #879, opened 2013-10-26
Fixing Digital Ocean VMs being ignored in delete map commands

JustinCarmony:
DO NOT MERGE RIGHT AWAY! PLEASE READ!

This is the simple quick fix, but might not be the right approach. I believe the apache cloud lib normalizes the vm state data so that if a VM is up and running it returns a state of "running." However, because the digital ocean API is used directly, it returns the state as "Active", hence it gets popped out of the list.

A better fix than what I have bellow is having the Digital Ocean driver normalize the value from active to running, but I Don't know if that would break anything for others. So hence I'm asking which approach we should do. :)

This pull request is in response to the issue #875

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -914,7 +914,7 @@ class Map(Cloud):
                 for vm_name, vm_details in vms.copy().iteritems():
                     if vm_details == 'Absent':
                         query_map[alias][driver].pop(vm_name)
-                    elif vm_details['state'].lower() != 'running':
+                    elif vm_details['state'].lower() != 'running' and vm_details['state'].lower() != 'active':
                         query_map[alias][driver].pop(vm_name)
                 if not query_map[alias][driver]:
                     query_map[alias].pop(driver)


techhat:
I'm inclined to agree with this fix, since it causes minimal changes to others that might be expecting `active` from this driver. But I'm going to sleep on it, and discuss it with @s0undt3ch in the morning, before we merge it.

saltcloud/cloud.py
```diff
@@ -914,7 +914,7 @@ def delete_map(self, query=None):
                 for vm_name, vm_details in vms.copy().iteritems():
                     if vm_details == 'Absent':
                         query_map[alias][driver].pop(vm_name)
-                    elif vm_details['state'].lower() != 'running':
+                    elif vm_details['state'].lower() != 'running' and vm_details['state'].lower() != 'active':
```
s0undt3ch, line 917: Minor syntax change:

``` python
elif vm_details['state'].lower() not in ('running', 'active'):
```

techhat:
Thanks @JustinCarmony!

PR #880, opened 2013-10-26
Use cElementTree if available

holmboe:
Same code as used in Salt.


techhat:
Thanks @holmboe. I'm going to hold off on this for two reasons. First, I've been lazy about converting this to use the same function that I built into salt:

https://github.com/saltstack/salt/blob/develop/salt/utils/xmlutil.py

Second, we're planning to merge salt-cloud into salt in the next couple of weeks, and at that point it can take advantage of salt's Python. compatibility layer.

s0undt3ch:
And salt-cloud depends on salt so we can import from salt's `_compat`, no?

techhat:
Agreed.

holmboe:
Ah, no stress. I had picked up on the planned/proposed salt-cloud -> salt merge in a comment somewhere so I had that in mind actually.

techhat:
Thanks for the pull, and for being so understanding, @holmboe!

holmboe:
:thumbsup:

PR #882, opened 2013-10-28
Added ability to create persistent spot instances on EC2

lusid:
EC2 supports the ability to create one-time or persistent spot instances. The difference is that persistent instances will automatically start back up again if they are shut down by Amazon due to price adjustment.

Example:

test-server:
  spot_config:
    type: persistent
    spot_price: 1

It still defaults to a "one-time" instance if "type" is not specified so it should function the same as before unless someone specifically wants a persistent spot.


techhat:
@lusid, you're my favorite person today. You rock, man.

PR #885, opened 2013-10-30
Remove duplicate entries in documentation

holmboe:
The two release notes are also included in _topics/releases/index_ thus they
appeared twice in the documentation.


holmboe:
Ah gotta love CI, https://salt-cloud.readthedocs.org/en/latest/contents.html is already updated.

techhat:
Wow, that was fast!

PR #893, opened 2013-11-05
Allow 'minion' to be absent in cloud.profiles entries - fixes #701

matthewryanscott:
Prior to this patch, you'd get the following traceback if you didn't specify a 'minion' key in an entry in /etc/salt/cloud.profiles:

$ sudo salt-cloud -m /etc/salt/cloud.map
[INFO    ] Applying map from '/etc/salt/cloud.map'.
[ERROR   ] There was a query error: 'minion'
  File "/opt/salt-cloud/src/salt-cloud/saltcloud/cli.py", line 273, in run
    dmap = mapper.map_data()
  File "/opt/salt-cloud/src/salt-cloud/saltcloud/cloud.py", line 1102, in map_data
    if 'grains' in nodedata['minion']:
KeyError: 'minion'

### DIFF
--- a/saltcloud/cloud.py
+++ b/saltcloud/cloud.py
@@ -1099,7 +1099,7 @@ class Map(Cloud):
                 # merge minion grains from map file
                 if 'minion' in overrides:
                     if 'grains' in overrides['minion']:
-                        if 'grains' in nodedata['minion']:
+                        if 'grains' in nodedata.get('minion', {}):
                             nodedata['minion']['grains'].update(
                                 overrides['minion']['grains']
                             )


techhat:
@s0undt3ch, what are your thoughts on this?

s0undt3ch:
Still need to test the provided configuration to see if I reproduce. This fixes the problem but we need to make sure it's not something deeper in the code.

techhat:
@gldnspud, I'm sorry that we didn't get this addressed before we got salt-cloud merged into Salt. Even though it is no longer physically possible to merge code into Salt Cloud from this repo, I'm going to leave this in until we can get it addressed in the new salt.cloud location. If we do end up using this change, we'll be sure to credit you appropriately, of course.

PR #895, opened 2013-11-07
Fix Issue #894, Make cli action option works with map file.

wangqiang8511:
Add get_vmnames_by_action in cloud.py Map, used for getting available names from map file, call get_vmnames_by_action when use action option in cli with map file.

After this patch, 

``` bash
salt-cloud -a start -m /etc/salt/maps/example.map
salt-cloud -a stop -m /etc/salt/maps/example.map
```

can get the correct available vm names from map file and work correctly. For "start" action only vms with state "stopped" are returned. For "stop" and "reboot" actions, vms with state "running" or "active" are returned.


techhat:
Thanks @wangqiang8511!

PR #899, opened 2013-11-27
Update the salt bootstrap script to a writeable path.

s0undt3ch:


s0undt3ch:
Still in time for 0.8.11?

techhat:
Ah, thanks for backporting this.

PR #901, opened 2013-12-12
Recommend reactor as method to initiate provisioning

akoumjian:
Since the startup_action config seems buggy at best, it made more sense to use salt's Reactor system as a way to provisioning a newly setup minion.


thatch45:
Since we merged in salt-cloud to salt itself we need to get these docs in there as well

akoumjian:
_is so out of the loop, proceeds to watch the past 10 salt airs_

thatch45:
:)

inthecloud247:
lol

PR #915, opened 2014-09-16
Update default for wait_for_ip_timeout

awestendorf:
As of version `2014.1.7`, it appears that `wait_for_ip_timeout` defaults to 10 minutes.


techhat:
Really? It used to be 20 minutes (because of IBM SCE, which has been replaced with SoftLayer), but then we moved it down to 5.

But I have no reason to change it from 10, so this looks good to me.

PR #920, opened 2014-12-29
add note that salt-cloud was merged into salt.

twellspring:
In case someone hits this page directly from a google search, added a note to say that salt-cloud is now part of salt proper.


techhat:
Hi @twellspring. We've tried to avoid changing this repo since the merge, and so our documentation regarding it was placed in the [readme](https://github.com/saltstack/salt-cloud/blob/develop/README.rst), so that it always appears on the front page of the repo.

rallytime:
I am closing this PR since we are no longer accepting changes in this repo. Thanks all!

